PR #19, opened 2017-08-20
Support AirQualityObserved

taliaga:
Simple support for Arrays and StucturedValues attributes, used for example in AirQualityObserved Model.

Plus, typo fix of "geo:json"


taliaga:
Simple support for Arrays and StucturedValues attributes, used for example in AirQualityObserved Model.

Issue #25, opened 2017-09-05
Add volumes to default recipe


taliaga:
Add docker volumes to recipes mentioned in the documentation.

PR #26, opened 2017-09-27
Add volumes to mongo and crate in Grafana experiment

taliaga:
Luckily as suspected, with stack deploy and volumes "not external", the volume is not re-created when existing.

I'm adding this in this compose because it's currently being used as "reference" by some people testing QL and they've requested persistence.

### DIFF
--- a/experiments/grafana/docker-compose.yml
+++ b/experiments/grafana/docker-compose.yml
@@ -19,13 +19,6 @@ services:
     image: mongo:3.2
     ports:
       - "27017:27017"
-    deploy:
-      mode: global
-      restart_policy:
-        condition: on-failure
-      update_config:
-        parallelism: 1
-        delay: 1m30s
     volumes:
       - mongodata:/data/db
 
@@ -48,13 +41,6 @@ services:
       # Transport protocol
       - "4300:4300"
     command: -Ccluster.name=democluster -Chttp.cors.enabled=true -Chttp.cors.allow-origin="*"
-    deploy:
-      mode: global
-      restart_policy:
-        condition: on-failure
-      update_config:
-        parallelism: 1
-        delay: 1m30s
     volumes:
       - cratedata:/data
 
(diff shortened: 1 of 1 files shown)


taliaga:
Fixes  #25

experiments/grafana/docker-compose.yml
```diff
@@ -19,6 +19,15 @@ services:
     image: mongo:3.2
     ports:
       - "27017:27017"
+    deploy:
+      mode: global
```
chicco785: if this is to be used with docker compose, deploy configuration does not make sense.
if this is to be used with docker stack, this configuration, using global will create an instance on each node, but the instances won't be in synch

experiments/grafana/docker-compose.yml
```diff
@@ -39,6 +48,15 @@ services:
       # Transport protocol
       - "4300:4300"
     command: -Ccluster.name=democluster -Chttp.cors.enabled=true -Chttp.cors.allow-origin="*"
+    deploy:
```
chicco785: if this is to be used with docker compose, deploy configuration does not make sense.
if this is to be used with docker stack, this configuration, using global will create an instance on each node, but the instances won't be in synch

taliaga: Right, you've got a point. It's currently being used with compose, so the deploy part is ignored. Adding a volume made me automatically add the "one per node" constraint, but sure, this won't be either enough to make it work with stack deploy. For that I'll push the other recipe. I just didn't want to leave this one as it was without persistence.

Responding your questions, removing a stack would typically not remove the volume. Deploying a stack when the volume was previously created will use the previously created volume.

chicco785: then i recommend either to make the compose script using a deploy with replica 1 (so that i can be used for testing using `docker stack deploy`) or to remove the deploy part

taliaga: Ok, removed deploy part to avoid confusion

PR #27, opened 2017-09-28
Add partitioning experiment

taliaga:
First attempt to get an impression on partitioning strategy


taliaga:
I'd like to run them in fiware lab and see what we find :)

Issue #30, opened 2017-11-06
Attributes names loose capitalization


taliaga:
Crate has some (naming restrictions)[https://crate.io/docs/crate/reference/en/latest/sql/ddl/basics.html#naming-restrictions] that forces the tables and column names to be lowercase.

We need to persist the original entity name and attribute names (the skeleton of the entity) somewhere so that information is used when constructing the entities to reply data queries.

PR #32, opened 2017-11-09
Respect capitals

taliaga:
Introduce the metadata table to persist information of original entity attributes names. This same table will be used for different metadata associated to each table.


translators/crate.py
```diff
@@ -2,28 +2,56 @@
...
+
 # NGSI TYPES: Not properly documented so this might change. Based on experimenting with Orion.
+NGSI_ID = 'id'
+NGSI_TYPE = 'type'
+NGSI_TEXT = 'Text'
+NGSI_DATETIME = 'DateTime'
+
 # CRATE TYPES: https://crate.io/docs/reference/sql/data_types.html
 NGSI_TO_CRATE = {
     "Array": 'array(string)',  # TODO: Support numeric arrays
```
chicco785: do we have a discussion on data types in the documentation pointing to the respective documents?

https://crate.io/docs/crate/reference/en/latest/appendix/data-types.html

chicco785: regarding array, i think there is no way to do a fixed mapping to a type,
you should check what is the type in the array, and dynamically generate it.
it could be even an object in term of json. a simple alternative could be to use object.

translators/crate.py
```diff
@@ -58,44 +87,52 @@ def _get_isoformat(self, ms_since_epoch):
...
-            for k, v in zip(keys, r):
-                if k == self.TIME_INDEX_NAME:
+            for k, v in zip(col_names, r):
+                original_name, original_type = entity_attrs[k]
+
+                if original_name in (NGSI_TYPE, NGSI_ID):
+                    entity[original_name] = v
+
+                elif original_name == self.TIME_INDEX_NAME:
                     # TODO: This might not be valid NGSI. Should we include this time_index in the reply?.
```
chicco785: is the time index a data created by crate? in which case, you should not include it in the reply i believe

translators/tests/test_crate.py
```diff
@@ -120,14 +131,69 @@ def test_benchmark_extended(translator):
...
+    entities = translator.query()
+    assert len(entities) == 2
+
+    assert_ngsi_entity_equals(e2, entities[1])
+    # Note that old entity gets None for the new attribute
+    e['NewAttr'] = {'type': 'Text', 'value': None}
+    assert_ngsi_entity_equals(e, entities[0])
+
+
 # FIWARE DATA MODELS
```
chicco785: we need to introduce more tests, at least all the ones that now have a decent schema definition.

please include:
* https://github.com/Fiware/dataModels/tree/master/Alert
* https://github.com/Fiware/dataModels/tree/master/WasteManagement
* https://github.com/Fiware/dataModels/tree/master/Device
* https://github.com/Fiware/dataModels/tree/master/PointOfInterest
* https://github.com/Fiware/dataModels/tree/master/Transportation/Bike/BikeHireDockingStation

translators/crate.py
```diff
@@ -2,28 +2,56 @@
...
     "Boolean": 'boolean',
-    "DateTime": 'timestamp',
+    NGSI_DATETIME: 'timestamp',
     "Integer": 'long',
     "geo:json": 'geo_shape',
     "Number": 'float',
-    "Text": 'string',
+    NGSI_TEXT: 'string',
     "StructuredValue": 'object'
 }
```
chicco785: we should had too: 
geo:point -> geo_point

translators/tests/test_crate.py
taliaga: Created https://github.com/smartsdk/ngsi-timeseries-api/issues/33 to be done in a new PR

translators/crate.py
taliaga: it's created by quantumleap in the table.
In general I agree, it shouldn't be part of the reconstructed entity. Let's see this in the PR of the API with the implementation of the first query method.

translators/crate.py
taliaga: Good one, we have it in the code, not in the docs. I'll add now.

taliaga: Forget, we do have in the docs. Each type of the conversion is linked to the corresponding crate docs.

chicco785: what about link to NGSI documentation?

chicco785:
codeclimate is my fault :)

translators/crate.py
taliaga: There's no documentation of ngsi types as far as I know, so the table in the doc lists the types I've seen in Orion docs. I will include such thing as soon as we have it :)

taliaga: right, that's more or less what the TODO is about.

translators/crate.py
taliaga: Created #35

translators/crate.py
taliaga: Created #36

chicco785: if the object storage works, i would the simplest (and probably more performing) solution.

doc/manuals/user/index.md
```diff
@@ -108,6 +107,9 @@ However, you can already query the persisted data either directly interacting wi
 
 What you need to know in the mean time is that QuantumLeap will create one table per each entity type. Table names are formed with the "et" prefix plus the lowercase version of the entity type. I.e, if your entity type is *AirQualityObserved*, the corresponding table name will be *etairqualityobserved*. All created tables for now belong to the default "doc" schema.
 
+### Restrictions
+
+- You cannot have two entity types with the same name but different capitalization. E.g: `Room` and `room`. The same applies for attribute names of a given entity. I.e, attributes `Temperature` and `temperature` will be treated as the same.
```
chicco785: I think I already mentioned (over skype?), but I may be wrong. Here i would refer to the fact that NGSI guidelines for attributes recommend the adoption of camel case. Which means, if you have an attribute called Temperature, you are doing it wrong :)

taliaga: True, it may have been skype, not sure. Added clarification and link.

taliaga:
Closes #30

PR #38, opened 2017-11-23
Add traffic observer

taliaga:
This PR brings 3 things as consequence of the demo preparation:
1. A test for the translator with TrafficFlowObserved instance
  - Motivation: To have yet another "official model" in the demo example.
  - Future ideal: to have these tests pull models from somewhere in accordance to dataModels repository.
2. A relaxation on the entities types recognition.
  - Motivation: QL should not fail if user misstyped the type. Worst case, fallback to string.
  - Future ideal: to have QL attempt some guessings of the type so we can think of accepting "simplified notation" for entities.
3. The scripts in the dataModels experiment folder, used to generate random input of entities of those types.
  - Motivation: It's practical to have them at hand in the same folder
  - Future ideal: Replace these with something more aligned with the [device simulator](https://github.com/telefonicaid/fiware-device-simulator). Even though it's much more sophisticated than our current needs, there might be some common factor.


experiments/dataModels/air_quality_observer.py
```diff
@@ -0,0 +1,129 @@
...
+        [51.208948, 4.418556],
+        [51.217179, 4.341202],
+        [51.218305, 4.336690],
+    ]
+
+    for n in range(N_ENTITIES):
+        entity = {
+            "id": 'air_quality_observer_{}'.format(n),
+            "type": "AirQualityObserved",
+            "laneId": n % 2,
```
chicco785: laneId?

chicco785:
i would also do a little bit of linting on the modified files so us to make codeclimate more happy ;)

experiments/dataModels/air_quality_observer.py
taliaga: good catch.

taliaga:
From 46 down to 12. Some are to be ignored indeed (long lines due to links or markdown tables).
Others I have them in my todo list, specially re-writing the insert method which is looking more than ugly :p. In any case, thanks for bringing the attention to this.

PR #47, opened 2018-01-02
filter get entities - response with reduced size

juliozinga:
Add ?attrs=null
Change entity query: /v2/entities?attrs=null


chicco785:
Thx for your PR @taliaga will follow up with you!

It would be nice to know what are your plans for using quantumleap and your experience with it.

experiments/dataModels/clear_orion.py
```diff
@@ -33,7 +33,7 @@
     print("removed {} subscriptions.".format(len(subs)))
 
     # Clean entities
-    entities = requests.get('{}/v2/entities'.format(ORION_URL),
+    entities = requests.get('{}/v2/entities?attrs=null'.format(ORION_URL),
```
taliaga: good one

experiments/dataModels/subscribe_ql.py
```diff
@@ -15,8 +15,15 @@
...
+    if entity_type.startswith('Traffic'):
+        description = "traffic_flow_observed"
+    elif entity_type.startswith('Air'):
+        description = "air_quality_observed"
+    else:
+        description = "description"
+
     s = {
-        "description": "traffic_flow_observed",
+        "description": description,
```
taliaga: How about using the entity_type directly as the description (instead of the ifs)? I'd rather keep it simple, after all it's a simple experiment :)

taliaga:
Hi @juliozinga, thanks for your interest to contribute!

As a first thing I would suggest you update your branch with the latest version of this repo's master so that you actually are synced with the latest status of the code.

Regarding your changes, a couple of questions:
- To be honest, the use of `Fiware-Service` is not so clear to me. I know it's a string to be used when running Orion with `multiservice` flag (for multitenancy), and its meaning is scenario-specific. So, is it wrong to use `default`? Where did you get the empty string info from?
- For the rest see comments on code

PR #48, opened 2018-01-03
Fix master

taliaga:
Collect and return pytest status of course!

Having seen the [build fail](https://travis-ci.org/smartsdk/ngsi-timeseries-api/builds/324643506), proceeded to fix the typo.


chicco785:
still failing, which is weird, given that the push one passed.

Issue #14, opened 2017-07-20
Support multi-tenancy using service path concept of context broker


chicco785:
When we deal with different scenarios sharing a common infrastructure, but we do not want them to share access to data, we need to find a way to isolate the data.
A simple solution, as the one used in context broker vs mongodb, is to have one database for tenant (identified by the service-path in the header request).

taliaga:
Change DB_NAME = "ngsi-tsdb" to `default`, as suggested in https://github.com/smartsdk/ngsi-timeseries-api/pull/49

taliaga:
Fixed by https://github.com/smartsdk/ngsi-timeseries-api/pull/65

PR #49, opened 2018-01-04
Add api

taliaga:
First attempt to have a swagger-defined query api leveraging on connexion framework.

Discussions on quantumleap.yml are welcome


reporter/reporter.py
```diff
@@ -115,10 +121,81 @@ def notify():
...
-if __name__ == '__main__':
-    app.run(host=LOCAL, port=8668)
+def query_1T1E1A(attrName, entityId, type=None):
+    """
+    See /entities/{entityId}/attrs/{attrName} in API Specification
+    quantumleap.yml
+    """
+    DB_HOST = os.environ.get('CRATE_HOST', 'crate')
+    DB_PORT = 4200
+    DB_NAME = "ngsi-tsdb"
```
chicco785: i would use a name that later on it can be easily used for the multi tenancy without the need of updating it.

e.g. default

reporter/reporter.py
```diff
@@ -115,10 +121,81 @@ def notify():
...
+
+    from translators.crate import CrateTranslator
+    TIME_INDEX = CrateTranslator.TIME_INDEX_NAME
+    with CrateTranslator(DB_HOST, DB_PORT, DB_NAME) as trans:
+        attr_names = [TIME_INDEX, attrName]
+        vals = trans.query(attr_names=attr_names, entity_type=type,
+                           entity_id=entityId)
+    res = {
+        'data': {
```
chicco785: do we need the data wrapper?

reporter/reporter.py
```diff
@@ -115,10 +121,81 @@ def notify():
...
 
-    msg = "Notification successfully processed"
+    msg = 'Notification successfully processed'
     logger.info(msg)
     return msg
 
 
-if __name__ == '__main__':
-    app.run(host=LOCAL, port=8668)
+def query_1T1E1A(attrName, entityId, type=None):
```
chicco785: no filters?

taliaga: let's see if I can include any in this same PR. Any in particular you had in mind?

reporter/reporter.py
taliaga: good idea

reporter/reporter.py
taliaga: I thought this would be useful because the response will in many cases come with some metadata (eg, page number), so it's a good way to keep the pure data clean. It's analogous to the idea of "contextElement" in [STH](http://fiware-sth-comet.readthedocs.io/en/latest/raw-data-retrieval/index.html)

taliaga:
@chicco785 it is indeed committed, you may need to click "load diff" to see it in detail. [Here](https://github.com/smartsdk/ngsi-timeseries-api/pull/49/files#diff-2527136b5a7dc1a912bd13b04e99c143)

specification/quantumleap.yml
```diff
@@ -0,0 +1,500 @@
...
+          examples:
+            application/json:
+              {
+                "entityId": "Kitchen1",
+                "attrName": "temperature",
+                "index": [1510826981388, 1510826990438, 1510826999725],
+                "values": [24.1, 25.3, 26.7]
+              }
+
+  # TODO: Could we remove all these /value suffixed endpoints if we provide the 'options' query arg? (keyvalues or values)?
```
taliaga: ...or we keep both options

reporter/reporter.py
taliaga: in fact we can get some inspiration from things like [these](http://jsonapi.org/) and harmonise all responses.

reporter/reporter.py
chicco785: i am not thinking about their implementation, but it would be better if the method already foresee them.

specification/quantumleap.yml
```diff
@@ -0,0 +1,500 @@
...
+    enum: [count, sum, avg, min, max]  # These taken from ISO standard on SQL.
+    # Remember Crate supports many others https://crate.io/docs/crate/reference/en/latest/sql/aggregation.html
+    description: "Optional. The function to apply to the raw data filtered by the query parameters. If not given, the returned data are the same raw inserted data."
+  aggrPeriod:
+    in: query
+    name: aggrPeriod
+    type: string
+    enum: [year, month, day, hour, minute, second]
+    default: month
+    description: "Optional. If defined, the aggregation function will be applied N times, once for each period, and all those results will be considered for the result. For example, a query asking for the average temperature of an attribute will typically return 1 value. With an aggregationPeriod of day, you get the daily average of the temperature instead."
```
chicco785: if not the defined, shouldn't it be on the overall data retrieved?

specification/quantumleap.yml
chicco785: we should keep it

specification/quantumleap.yml
```diff
@@ -0,0 +1,500 @@
...
+                "index": [1510826981388, 1510826990438, 1510826999725],
+                "values": [24.1, 25.3, 26.7]
+              }
+
+  # TODO: Could we remove all these /value suffixed endpoints if we provide the 'options' query arg? (keyvalues or values)?
+  /entities/{entityId}/attrs/{attrName}/value:
+    get:
+      operationId: reporter.reporter.query_1T1E1A_value
+      summary: "History of an attribute (values only) of a given entity instance."
+      description: "Similar to the previous, but focusing on the values regardles of the metadata."
```
chicco785: regardles -> regardless

specification/quantumleap.yml
```diff
@@ -0,0 +1,500 @@
...
+        - $ref: '#/parameters/attrNameInPath'
+        # Query attrs to be refined...
+        - $ref: '#/parameters/type'
+        - $ref: '#/parameters/aggrMethod'
+        - $ref: '#/parameters/aggrPeriod'
+        - $ref: '#/parameters/options'
+        - $ref: '#/parameters/from'
+        - $ref: '#/parameters/to'
+        - $ref: '#/parameters/lastN'
+        - $ref: '#/parameters/limit'
```
chicco785: offset?

specification/quantumleap.yml
```diff
@@ -0,0 +1,500 @@
...
+      operationId: reporter.reporter.query_1T1E1A_value
+      summary: "History of an attribute (values only) of a given entity instance."
+      description: "Similar to the previous, but focusing on the values regardles of the metadata."
+      tags:
+        - queries
+      parameters:
+        - $ref: '#/parameters/entityIdInPath'
+        - $ref: '#/parameters/attrNameInPath'
+        # Query attrs to be refined...
+        - $ref: '#/parameters/type'
```
chicco785: probably:
 +        - $ref: '#/parameters/aggrMethod'
 +        - $ref: '#/parameters/aggrPeriod'
 +        - $ref: '#/parameters/from'
 +        - $ref: '#/parameters/to'
 +        - $ref: '#/parameters/lastN'
 +        - $ref: '#/parameters/limit'
 +        - $ref: '#/parameters/offset'

specification/quantumleap.yml
```diff
@@ -0,0 +1,500 @@
...
+                description: "Array of values of the selected attribute, in the same corresponding order of the 'index' array."
+          examples:
+            application/json:
+              {
+                "entityId": "Kitchen1",
+                "attrName": "temperature",
+                "index": [1510826981388, 1510826990438, 1510826999725],
+                "values": [24.1, 25.3, 26.7]
+              }
+
```
chicco785: what happens when using aggregates?

specification/quantumleap.yml
```diff
@@ -0,0 +1,500 @@
...
+                description: "Array of values of the selected attribute, in the same corresponding order of the 'index' array."
+          examples:
+            application/json:
+              {
+                "id": "Kitchen1",
+                "attrName": "temperature",
+                "index": [1510826981388, 1510826990438, 1510826999725],
+                "values": [24.1, 25.3, 26.7]
+              }
+
```
chicco785: what happens with two attributes or more attributes?

a good approach could be (and when for a time index there is not value for an attribute, we set it to null):

```
{
     "id": "Kitchen1",
     "index": [1510826981388, 1510826990438, 1510826999725],
     "attributes": [
          {
             "attrName": "temperature",
             "values": [24.1, 25.3, 26.7]
          },
          {
             "attrName": "pressure",
             "values": [1.01, 0.9, 1.02]
          },
     ]
}
```

specification/quantumleap.yml
```diff
@@ -0,0 +1,500 @@
...
+        - queries
+      parameters:
+        - $ref: '#/parameters/entityTypeInPath'
+        - $ref: '#/parameters/attrNameInPath'
+        # Query attrs to be refined...
+        - $ref: '#/parameters/id'
+      responses:
+        200:
+          description: OK
+          # Response types to be refined...
```
chicco785: what if two entities have the same index? should we keep track of the entities?

in case the reply could look like (and in case for entity on a given index there is not data, we use null):
```
{
     "entityType": "Room",
     "attrName": "temperature",
     "index": [1510826981388, 1510826990438, 1510826999725],
     "entities": [
          {
             "entityId": "Kitchen1",
             "values": [24.1, 25.3, 26.7]
          },
          {
            "entityId": "Room2",
             "values": [23.01, 24.9, 23.02]
          },
     ]
}
```
or it may be better to use separate indexes:

in case the reply could look like (and in case for entity on a given index there is not data, we use null):
```
{
     "entityType": "Room",
     "attrName": "temperature",
     "entities": [
          {
             "entityId": "Kitchen1",
             "index": [1510826981388, 1510826990438, 1510826999725],
             "values": [24.1, 25.3, 26.7]
          },
          {
            "entityId": "Room2",
            "index": [1510826981388, 1510826990438, 1510826999725],
             "values": [23.01, 24.9, 23.02]
          },
     ]
}
```

we should be able to specify if the aggregate is global or per entity.

specification/quantumleap.yml
```diff
@@ -0,0 +1,500 @@
...
+        - queries
+      parameters:
+        - $ref: '#/parameters/attrNameInPath'
+        # Query attrs to be refined...
+        - $ref: '#/parameters/type'
+        - $ref: '#/parameters/id'
+      responses:
+        200:
+          description: OK
+          # Response types to be refined...
```
chicco785:
```
{
     "attrName": "temperature",
     "entities": [
          {
             "entityId": "Kitchen1",
             "entityType": "Room",
             "index": [1510826981388, 1510826990438, 1510826999725],
             "values": [24.1, 25.3, 26.7]
          },
          {
            "entityId": "Station1",
            "entityType": "WeatherStation",
            "index": [1510826981388, 1510826990438, 1510826999725],
             "values": [23.01, 24.9, 23.02]
          },
     ]
}
```

in this case aggregation could be global, or per entityType

specification/quantumleap.yml
```diff
@@ -0,0 +1,500 @@
...
+        - queries
+      parameters:
+        - $ref: '#/parameters/entityIdInPath'
+        - $ref: '#/parameters/attrNameInPath'
+        # Query attrs to be refined...
+        - $ref: '#/parameters/type'
+      responses:
+        200:
+          description: OK
+          # Response types to be refined...
```
chicco785:
```
{
  "index": [1510826981388, 1510826990438, 1510826999725],
  "values": [24.1, 25.3, 26.7]
}
```

specification/quantumleap.yml
```diff
@@ -0,0 +1,500 @@
...
+        - queries
+      parameters:
+        - $ref: '#/parameters/entityIdInPath'
+        # Query attrs to be refined...
+        - $ref: '#/parameters/type'
+        - $ref: '#/parameters/attrs'
+      responses:
+        200:
+          description: OK
+          # Response types to be refined...
```
chicco785: a good approach (are array of arrays possible in json?) could be (and when for a time index there is not value for an attribute, we set it to null):

```
{
     "index": [1510826981388, 1510826990438, 1510826999725],
     "values": [
          [24.1, 25.3, 26.7],
          [1.01, 0.9, 1.02]
     ]
}
```

chicco785:
i missed the file :) good spot on the "Load diff"

reporter/reporter.py
taliaga: One thing though, unless I misunderstood your idea, this change would break existing usages as of now (I think of the mexican apps, they'll need to rename their dbs). I believe it might be better to change to 'default' when implementing the multitenancy.

specification/quantumleap.yml
taliaga: I'm afraid I don't understand the question, could you rephrase?

specification/quantumleap.yml
taliaga: Right, `offset` is missing.
Just to recap (for other reviewers also), using `offset` + `limit` would be like using `from`+`to`, just that the former pair works on indexes (positional numbers in response) and the latest with data (like from 01/01/2017 to today).
`lastN` will be later applied to any query result before being returned, it's like a final slicing.

specification/quantumleap.yml
taliaga: Arrays of arrays are valid AFAIK, but we'll probably need a "map" there to tell which array corresponds to which attribute (because I don't recall reading about default explicit attributes order)

reporter/reporter.py
chicco785: 1) let's make it configurable.
2) let's provide a guide on how to rename the database in case of upgrade

specification/quantumleap.yml
chicco785: let's suppose that i don't specific an aggregation period. the aggregate should be compute on the overall time frame retrieved.

specification/quantumleap.yml
chicco785: (y)

specification/quantumleap.yml
chicco785: can't it be a semantic we define, for example based on the order of attributes in the query?

?attrs=temperature,pressure&...

specification/quantumleap.yml
taliaga: This brings us to an interesting point to decide, the UTC in milliseconds vs ISO8601.
Technically the UTC in milliseconds is more "ideal", but looking at the context, if the user/notification is already bringing data in in ISO8601, then I believe the response should be consistent and reply the same format. If course, if the user inserted UTC milliseconds, the response will be like your example above. 
What do you think?

specification/quantumleap.yml
taliaga: I was also thinking in this approach as the default, being a unique device, chances are that it pushes all data at once (so different measurements sharing the same timeset).

specification/quantumleap.yml
chicco785: probably it could be a parameter, e.g. indexFormat and possible values are UTC and ISO8601 where ISO8601 is the default

chicco785: btw how the index are stored in crate?

specification/quantumleap.yml
chicco785: agree, the problem comes with different entities, where the indexes may be totally different

chicco785: see comment on row 381

reporter/reporter.py
taliaga: Sure, I was not concerned because the definite list of params was still to be defined, adding them is trivial though better at the end. Added params from the current definition and will update accordingly :)

taliaga:
Added some comments in the `quantumleap.yml` to trigger some discussions and tackled most of the open issues.

If this is getting messy to review, let me know and I can open a new PR or do something about it!

After next iteration maybe we can include Jose and/or German as reviewers of the specification.

reporter/reporter.py
taliaga: since it's not related to API I'll do this later. Added reminder in #14

Issue #37, opened 2017-11-23
geo:json for everyone


taliaga:
As suggested by @chicco785 , it would be interesting to have, for the entities expressing location in a non geo-json format (say, with country codes and addresses), an under-the-hood transformation to an equivalent geo-json structure, so that geo-queries can be made across different entities expressing locations in different ways.

I open this issue to keep track of the discussion of this idea.

chicco785:
some pointers that could be useful:
* https://github.com/DenisCarriere/geocoder
* https://github.com/datasets/geo-nuts-administrative-boundaries/tree/master/data

let's suppose that for some data we can do this in a easier way

for example, for country names and country codes, we could create a separate table
with the associated geo:json, and when running ngsi geo queries, do a merge on the field in the background.

for things like street address, we would definitively need a geocoding service but then, instead of querying the service every time i would cache the data

in relation to NGSI Data Models, when you import the Location-Commons definition in a schema, that schema will have geo:json and/or address, address object includes properties that allows us for some resolution, so if you have only addressCountry, we would do a resolution based on table merge (see above), if it has also an addressStreet, we would use an external service to obtain a geo:point
and cache it.
similar approach for entities with only these fields would be possible:
 "addressLocality": {
              "type": "string"
            },
"addressRegion": {
              "type": "string"
},

taliaga:
First version of this already implemented.

PR #56, opened 2018-01-25
Fix lat,long from osm

taliaga:


greengoaxe:
All tests passed.

taliaga:
#37

PR #58, opened 2018-02-27
adjust subscription json format

juliozinga:
short adjust

### DIFF
--- a/utils/common.py
+++ b/utils/common.py
@@ -131,6 +131,11 @@ def iter_entity_attrs(entity):
 
 
 def create_simple_subscription(notify_url):
+    """
+    Create a string subscription using NGSIv2 version *For testing only*
+    :param notify_url: reference to the registration
+    :return: string subscription using NGSIv2 version
+    """
     subscription = {
         "description": "Test subscription",
         "subject": {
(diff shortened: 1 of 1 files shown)


utils/common.py
```diff
@@ -131,6 +131,11 @@ def iter_entity_attrs(entity):
 
 
 def create_simple_subscription(notify_url):
+    """
+    Create a string subscription using NGSIv2 version *For testing only*
+    :param notify_url: reference to the registration
+    :return: string subscription using NGSIv2 version
```
taliaga: Hi @juliozinga

First of all, nice to see you back around here. I've been very careless with the documentation of these testing artifacts, I thought them as very disposable pieces of code, but if you are willing to fix these it means they're not so ignored after all :)

Note both of these test methods return a python dict, not a string.

Also, in python, I would suggest always keeping the last `,` in the lists. It's harmless and it saves time when you extend the lists with new elements (who never forgot to add the comma in the item above when extending a list). Does it make sense to you?

juliozinga: I don't understand much about that part of the code. I figured it would be transformed into a generic method later. In this sense I thought it should be commented. But if the goal is just test, it can be removed to another file so that it is explicit.
I'll be following more closely in the next few days. Greetings!

PR #59, opened 2018-03-01
Add api2

taliaga:


reporter/reporter.py
```diff
@@ -157,41 +150,6 @@ def add_geodata(entity):
...
-
-# Before implementing the rest of these methods, move query methods to a
-# different module.
-
-def query_1T1E1A_value():
-    raise NotImplementedError
-
-
 def query_1T1ENA():
     raise NotImplementedError
```
chicco785: i saw code for query_1T1ENA is this still valid?

taliaga: good catch, removed.

Issue #22, opened 2017-08-22
Suggestions for refactoring


chicco785:
looking for tests, i would expect a main folder ```tests``` and then inside it the test folders for each module, rather than the other way around. but probably is just a a matter of taste.

also, i would move the source of the module in ```src``` folder, so to better structure the repository

for the client, we have a full client in python (https://github.com/smartsdk/ngsi-sdk-python), so it was not needed to develop one just for the purpose of the tests... use your own dog food.

it looks also weird that the folder is named client, when it is an orion client and not the quantum leap one.

it looks quite weird as well to have certain files in the root:
conftest.py
run.sh (this run the tests right? the name does not point to that)

i also suggest to have a Docker folder that include the Dockerfile, but also a default example composition. more complex one can be in the recipe repository.

chicco785:
also the benchmark part should not be in functional part of the code, a separate folder would help in navigating the code.

taliaga:
Totally agree with all the comments. This need for a refactoring has been a known. 

Moreover, this repo was initially supposed to host only API controllers (whose development is on hold until we have at least a couple of iterations on the prototype and better understanding of the whole). Instead, this is hosting a bit of all the pieces we've put together so far.

The tests files location is a matter of taste AFAIK. To me it's easier this way because otherwise, to keep things controllable, you need to replicate the modules tree structure in src, leading to possible missplacings. I may be missing some benefits of the separate test folder though.

For the records of this task, another thing to be done is to split the requiremets.txt into test-only and prod dependencies. This goes in the line of making docker image smaller.

The benchmark we can move for sure. Which files and at which level would you move it? Note it's in the "functional part" to keep it as "reproduceable" as possible. The benchmark functions are part of the test suite and the plots are generated out of it.

chicco785:
> The tests files location is a matter of taste AFAIK. To me it's easier this way because otherwise, to keep things controllable, you need to replicate the modules tree structure in src, leading to possible missplacings. I may be missing some benefits of the separate test folder though.

Code readability :) separating tests seems to be also a python coding recommendation: http://docs.python-guide.org/en/latest/writing/structure/

PR #77, opened 2018-07-30
Move QL modules to src folder

taliaga:
Part two of cleanup!


taliaga:
Also covering points of https://github.com/smartsdk/ngsi-timeseries-api/issues/22

PR #83, opened 2018-08-23
Implement query 1TNE1A

taliaga:
This means, endpoints:
- /types/{entityType}/attrs/{attrName}
- /types/{entityType}/attrs/{attrName}/value

### DIFF
--- a/specification/quantumleap.yml
+++ b/specification/quantumleap.yml
@@ -215,42 +215,48 @@ parameters:
     name: aggrPeriod
     type: string
     enum: [year, month, day, hour, minute, second]
-    description: "Optional. In not defined, the aggregation will apply to all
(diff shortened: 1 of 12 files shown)


Cerfoglg:
@taliaga Codeclimate always seems to complain about very necessary ways of writing the code in this case (number of function arguments, ...)

taliaga:
@Cerfoglg That's True. Sometimes it also points out some things which are not worth following 100% because it'd be too expensive. I try to keep a balance, but other people's observations are critical.

From the list of issues, which would you consider the most important to tackle?

B.t.w, I did a quick search and couldn't find a way to avoid those of the number of params for the methods that control the conexion (API framework) calls. It's like the method should be prepared to handle any/all of the params a user could use in a REST call (both in path and in query).

PR #88, opened 2018-10-12
Add health endpoint

### LINKED ISSUE #68: Add status endpoint
Add some form of status endpoint, which could be used in liveness/readiness Probes that reports on the status of complementing services (crate, maybe grafana and redis)

taliaga:
Closes  #68

### DIFF
--- a/src/reporter/health.py
+++ b/src/reporter/health.py
@@ -1,7 +1,7 @@
 import os
 
 
-def check_crateDB():
+def check_crate():
     """
     crateDB is the default backend of QuantumLeap, so it is required by
     default.
@@ -41,6 +41,17 @@ def check_geocoder():
     return health
 
 
+def _get_http_code(res):
+    if res['status'] == 'pass':
+        code = 200
+    elif res['status'] == 'warn':
+        code = 207
+    else:
+        assert res['status'] == 'fail'
+        code = 424
+    return code
+
+
 def get_health():
     """
     Return status of QuantumLeap service, taking into account status of the
@@ -54,7 +65,7 @@ def get_health():
     res = {}
 
     # Check crateDB (critical)
-    health = check_crateDB()
+    health = check_crate()
     res['status'] = health['status']
     if health['status'] != 'pass':
         res.setdefault('details', {})['crateDB'] = health
@@ -74,12 +85,5 @@ def get_health():
             res['status'] = 'warn'
 
     # Determine HTTP code
-    if res['status'] == 'pass':
-        code = 200
-    elif res['status'] == 'warn':
-        code = 207
-    else:
-        assert res['status'] == 'fail'
-        code = 424
-
+    code = _get_http_code(res)
     return res, code, {'Cache-Control': 'max-age=180'}
--- a/src/reporter/health.py
+++ b/src/reporter/health.py
@@ -1,0 +1,85 @@
+import os
+
+
+def check_crateDB():
+    """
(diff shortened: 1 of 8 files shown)


src/reporter/reporter.py
```diff
@@ -170,55 +170,55 @@ def query_1TNENA():
...
+    return r, 501
 
 
 def config():
     r = {
         "error": "Not Implemented",
         "description": "This API method is not yet implemented."
     }
-    return r, 400
+    return r, 501
```
chicco785: +1

src/reporter/health.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+def _get_http_code(res):
+    if res['status'] == 'pass':
+        code = 200
+    elif res['status'] == 'warn':
+        code = 207
+    else:
+        assert res['status'] == 'fail'
+        code = 424
```
chicco785: shouldn't be 503? (service not available)

src/reporter/health.py
```diff
@@ -0,0 +1,89 @@
...
+    from geocoding import geocoding
+    health = geocoding.get_health()
+    return health
+
+
+def _get_http_code(res):
+    if res['status'] == 'pass':
+        code = 200
+    elif res['status'] == 'warn':
+        code = 207
```
chicco785: never seen this used, what's the case with ELK api? that maybe a good reference

src/reporter/health.py
taliaga: I'm not sure really, it's kind of a paradox. My reasoning was: I would expect from the server a 503 if the `/health` service was down so I can't know what's the health. But if it's working and telling you something's wrong with the dependencies, then we could use [424 Failed Dependency](https://httpstatuses.com/424)

src/reporter/health.py
taliaga: good idea, let me see...

taliaga: Elasticserach returns 200 even for a cluster with "yellow" status, so it may be better to stick with 200. Besides, 207 might require something else to properly implement multistatus.

src/reporter/health.py
chicco785: what's elk giving back?

src/reporter/health.py
chicco785: then let's use 200, I think 207 is not much adopted.

src/reporter/health.py
taliaga: Well, not sure, because it depends on the error type. I see you could get http 200 even with some red status (e.g. with one broken shard,  in which case it'd still return partial results).

taliaga: Since there's no standard, I thought of following this draft: https://inadarei.github.io/rfc-healthcheck/
which suggests that fail status should return codes in the 4XX/5XX range.

src/reporter/health.py
taliaga: ok!

src/reporter/health.py
chicco785: wait a minute, if the service is partially working, you don't have a fail.

taliaga: I took it from here: https://www.elastic.co/guide/en/elasticsearch/reference/current/_cluster_health.html

taliaga: OK, let's use 503. On a second thought, this looks more reasonable/common.

chicco785:
@taliaga to keep clean the merge, maybe is better if we move to "squash and marge"

PR #92, opened 2018-10-18
Alter maintained badge

jason-fox:
Update Core badge to use nexus CDN and amend status badge, this fulfils the maintained badge TSC requirement.


jason-fox:
It is broken for me at the moment too. Looks like the `fiware.org` website is down temporarily

* https://downforeveryoneorjustme.com/fiware.org

(And the idea was to pass badge maintainence over to FIWARE and for the badges to be more stable than `shields.io` [sigh])

Fortunately, once `fiware.org` **is** back up, `githib.com` will **cache** the images, and they have a longer timeout than `shields.io`

jason-fox:
@taliaga  - it looks like the nexus server is back up now.

PR #95, opened 2018-10-25
Ref aggr period

taliaga:
Adjusting return of `query` method of crate's translator to be more flexible for having different indexes.

### DIFF
--- a/specification/quantumleap.yml
+++ b/specification/quantumleap.yml
@@ -1,7 +1,7 @@
 swagger: '2.0'  # For 3.0 see (https://github.com/zalando/connexion/issues/420)
 info:
   title: "QuantumLeap API"
-  version: "0.1.1"
+  version: "0.1.2"
 host: "localhost:8668"  # it'll run in the same container, hence localhost.
 basePath: /v2
 produces:
@@ -215,16 +215,14 @@ parameters:
     name: aggrPeriod
     type: string
     enum: [year, month, day, hour, minute, second]
-    description: "(This parameter is not yet supported) Optional. If not
-    defined, the aggregation will apply to all the values contained in the
-    search result.
-    If defined, the aggregation function will instead be applied N times,
-    once for each period, and all those results will be considered for the
-    result. For example, a query asking for the average temperature of an
-    attribute will typically return 1 value.
-    However, with an aggregationPeriod of day, you get the daily average of
-    the temperature instead (more than one value assuming you had
-    measurements across many days within the scope of your search result)"
+    description: "Optional. If not defined, the aggregation will apply to all
+     the values contained in the search result. If defined, the aggregation
+     function will instead be applied N times, once for each period, and all
+     those results will be considered for the response. For example, a query
+     asking for the average temperature of an attribute will typically return
+      1 value. However, with an aggregationPeriod of day, you get the daily
+      average of the temperature instead (more than one value assuming you
+      had measurements across many days within the scope of your search result)"
   aggrScope:
     in: query
     name: aggrScope
@@ -566,6 +564,7 @@ paths:
             application/json:
               {
                 "data": {
+                    # Return type!?
                     "entityId": "Kitchen1",
                     "attrName": "temperature",
                     "index": ["2018-01-05T15:44:34", "2018-01-06T15:44:59",
@@ -696,7 +695,8 @@ paths:
               {
                 "data": {
                   "entityId": "Kitchen1",
-                  "index": ["1510826981388", "1510826990438", "1510826999725"],
+                  "index": ["2018-01-05T15:44:34", "2018-01-06T15:44:59",
+                            "2018-01-07T15:44:59"],
                   "attributes": [
                     {
                       "attrName": "temperature",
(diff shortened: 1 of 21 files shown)


specification/quantumleap.yml
```diff
@@ -566,6 +564,7 @@ paths:
             application/json:
               {
                 "data": {
+                    # Return type!?
```
chicco785: what's the problem here?

specification/quantumleap.yml
```diff
@@ -696,7 +695,8 @@ paths:
               {
                 "data": {
                   "entityId": "Kitchen1",
-                  "index": ["1510826981388", "1510826990438", "1510826999725"],
+                  "index": ["2018-01-05T15:44:34", "2018-01-06T15:44:59",
```
chicco785: shouldn't have also the time zone reference?

src/reporter/query_1T1E1A.py
```diff
@@ -57,16 +57,24 @@ def query_1T1E1A(attr_name,   # In Path
         return msg, 500
 
     if entities:
+        if len(entities) > 1:
+            import warnings
+            warnings.warn("Not expecting more than one result for a 1T1E1A.")
+
         if aggr_method:
```
chicco785: may be good to have a warning or error if aggr_period is used but not aggr_method

src/reporter/query_1T1ENA.py
```diff
@@ -57,25 +57,29 @@ def query_1T1ENA(entity_id,   # In Path
         return msg, 500
 
     if entities:
+        if len(entities) > 1:
+            import warnings
+            warnings.warn("Not expecting more than one result for a 1T1ENA.")
+
         if aggr_method:
-            index = []
+            if aggr_period:
```
chicco785: see above

src/translators/crate.py
```diff
@@ -475,13 +426,129 @@ def query(self,
...
+        composed of time steps of the original index of the attribute but
+        zeroing the less significant bits of time. For example, if there were
+        measurements in time 2018-04-03T08:15:15 and 2018-04-03T09:01:15, with
+        aggrPeriod = minute the new index will contain, at least, the steps
+        2018-04-03T08:15:00 and 2018-04-03T09:01:00 respectively.
+
+        :raises:
+        ValueError in case of misuse of the attributes.
+        UnsupportedOption for still-to-be-implemented features.
+        crate.DatabaseError in case of errors with CrateDB interaction.
```
chicco785: does this include time out? how do we handle very large queries? can we "segment"?

src/translators/crate.py
```diff
@@ -550,41 +615,109 @@ def query(self,
...
+
+        With aggrMethod and many specific ids requested, the array has
+        only one dict, but instead of an 'id' attribute it will have an
+        'ids'. This is the same case when user does not give a specific id at
+        all (no id => all ids).
+        returns [{
+                    'type': 'Room',
+                    'ids': ['Room1', 'Room2'],
+                    'temperature': {'values': [4,]},    # aggrMethod
+                    'index': [],                        # without aggrPeriod
```
chicco785: why not index?

specification/quantumleap.yml
taliaga: it was just a mental note, to think if we shouldn't maybe return the NGSI type of the attr as well.

specification/quantumleap.yml
taliaga: Good point. We stated in the docs that the returned time is always in UTC.
However, it may be nice to give this info back as well, wdyt? 

Also, adding some tests (see https://github.com/smartsdk/ngsi-timeseries-api/pull/95/commits/47356901b06a550b1f1ac7cdeb86abcd99cfbeb1) I've found an inconsistency, so I've adjusted #97 to cover these issues in a new PR.

src/reporter/query_1T1E1A.py
taliaga: Agree, and we have it :) 

https://github.com/smartsdk/ngsi-timeseries-api/blob/bc8d0bcb97b3e3d2709649cc34d7add9240a1d4c/src/reporter/tests/test_1T1E1A.py#L110

src/reporter/query_1T1ENA.py
taliaga: it's the same for all endpoints.

https://github.com/smartsdk/ngsi-timeseries-api/blob/bc8d0bcb97b3e3d2709649cc34d7add9240a1d4c/src/reporter/reporter.py#L286

src/translators/crate.py
taliaga: In spite of the name, `DatabaseError` is wrapping timeouts.
Queries are capped by QL and CrateDB at [10000](https://crate.io/docs/crate/reference/en/latest/general/dql/selects.html#limits).
As suggested by Crate folks, the user could make use of `offset` and `limit` to deal with paging at the moment. QL is exposing these two params, and there are simple unit-tests for both, but ideally this would need further testing.

src/translators/crate.py
taliaga: it's the case where user asks something like: 'average temperature for entities 1&2 from 2010 to 2011'. The response is just a number. There's no actual 'index'.

Another option is to make it an array of 2 steps: fromDate and toDate. That'd be the idea we mentioned in [this unimplemented method](https://github.com/smartsdk/ngsi-timeseries-api/blob/bc8d0bcb97b3e3d2709649cc34d7add9240a1d4c/specification/quantumleap.yml#L849). I'm still not convinced of its added value / effort ratio though.

chicco785: fair enough

specification/quantumleap.yml
chicco785: the this should be commented with TODO so you can scout it easily

Issue #69, opened 2018-05-11
Extend supported params in /subscribe endpoint


taliaga:
Add support for "id"  and "type" in addition to idPattern. See [orion docs](https://fiware-orion.readthedocs.io/en/master/user/walkthrough_apiv2/index.html#subscriptions)

If the user could otherwise specify two different `attrs` arrays, that'd be useful as well.

Throttling would be nice too.

PR #103, opened 2018-11-19
Extend supported params in /subscribe endpoint

c0c0n3:
This PR implements #69.

### DIFF
--- a/specification/quantumleap.yml
+++ b/specification/quantumleap.yml
@@ -490,6 +490,15 @@ paths:
           subscription, so as to persist historical data of entities of this
           type. ATTENTION: If not specified, all entity types will be tracked."
         - in: query
+          name: entityId
+          required: false
+          type: string
+          description: "Id of the entity to track. If specified, it takes precedence
+          over the idPattern parameter. If absent but an idPattern is given, then QL
+          will use the idPattern to determine which entities to track. If no idPattern
+          is given either, then QL will use the entityType if present. If not even an
+          entityType is given, then QL will fall back to tracking all entities."
+        - in: query
           name: idPattern
           required: false
           type: string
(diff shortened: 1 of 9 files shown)


c0c0n3:
Notes.
1. Refactoring. Moved subscription building logic in own module to make it unit-testable.
2. DSL. Implemented tiny EDSL to build subscriptions. Probably we should move this out of QL into a generic Orion client?
3. Key-value tree. The DSL is just syntactic sugar over this more generic component that lets you build key-value hierarchies where values may not be present (None) and tree paths get automatically pruned to exclude None values. This way, we can get rid of most of the conditional logic needed when building those dictionaries we eventually convert to JSON objects. Probably this doesn't belong in QL either, but I've parked it there for the time being...

c0c0n3:
looks like those CI failures have nothing to do with this PR's change set. In fact, after checking out master and starting from a clean slate (clean up env, rm images, etc.) I'm still getting those exact same errors in the geocoding tests...

chicco785:
> Notes.
> 
> 1. Refactoring. Moved subscription building logic in own module to make it unit-testable.

(y)

> 2. DSL. Implemented tiny EDSL to build subscriptions. Probably we should move this out of QL into a generic Orion client?

we had something autogenerated from OpenAPI at least for node.js. In the longer term, it would be better probably to follow that approach and embed it as a library.

> 3. Key-value tree. The DSL is just syntactic sugar over this more generic component that lets you build key-value hierarchies where values may not be present (None) and tree paths get automatically pruned to exclude None values. This way, we can get rid of most of the conditional logic needed when building those dictionaries we eventually convert to JSON objects. Probably this doesn't belong in QL either, but I've parked it there for the time being...

I am lost :)

src/reporter/subscription_builder.py
```diff
@@ -0,0 +1,29 @@
+from utils.subscription_dsl import *
+
+
+def build_subscription(quantumleap_url,
```
chicco785: not for now, but this does not work if the end point is protected. we need at least a way to pass a token over the header (and header parameter may be different depending on how the context broker end point is protected)

src/utils/kvt.py
```diff
@@ -0,0 +1,159 @@
+
+
+# lifted from:
+# - https://stackoverflow.com/questions/38987
+def merge_dicts(*dict_args):
```
chicco785: @taliaga this approach could be interesting (in general) for the creation of replies (if i understood the purpose). eventhough, i am not sure how much "manipulation" you are running on the json snippets.

src/reporter/subscription_builder.py
c0c0n3: Oh right, didn't know about that! But it's easy to change I reckon. Will need your help to understand better how it should work though.

src/utils/kvt.py
c0c0n3: yeah, you've hit the nail on the head. I'm not sure either it's worth our while unless we have to fiddle heavily with the json which I don't think we do at the moment. I was just testing the waters, we can ditch it if it doesn't turn out to be useful.

src/reporter/subscription_builder.py
chicco785: it's more a todo, not worries. in general we have to think about this auth aspects not only for consuming data, but also in case we interact with other APIs.

specification/quantumleap.yml
```diff
@@ -498,12 +507,36 @@ paths:
           specified type. This attribute is documented in orion, see
           https://fiware-orion.readthedocs.io/en/master/user/walkthrough_apiv2/index.html#subscriptions"
         - in: query
-          name: attributes
+          name: observedAttributes
```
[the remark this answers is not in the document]
c0c0n3: ...done, see commit 23ec9a993a077bdfc2768834f792f8aa13376ccd

Issue #113, opened 2018-11-27
check if docker image can be optimised 


chicco785:
https://blog.realkinetic.com/building-minimal-docker-containers-for-python-applications-37d0272c52f3

c0c0n3:
closed by #116

PR #116, opened 2018-11-29
optmize docker image (issue #113)

chicco785:
#113
now 170mb


chicco785:
for whatever reason certifi module is not available, so at run time it fails

c0c0n3:
re-rested locally. seems to work flawlessly. closes #113

PR #130, opened 2019-01-03
Avoid urls to old API versions

taliaga:
Signed-off-by: Tomas Aliaga <tomas.aliaga@gmail.com>


c0c0n3:
It looks like the build is failing because of an issue with (possibly) bad data returned by OSM. I don't think there's anything we can do about it at this stage, should we go ahead and merge anyway?

taliaga:
Let me see if I can make it less flaky trying to use some other points of the globe. Otherwise we'll have to skip the test overall.

Issue #128, opened 2018-12-27
attributes names and entity type containing '-' are not accepted by crateDB


agaldemas:
If an NGSI entity contains '-' in the type or in an attribute name, crateDB return an error and doesn't save it.
the table name composed with prefix and type should be quoted also attributes should be quoted like:
"my-attribute", then CrateDB accept it.
if this is done in every case the problem should be fixed.

agaldemas:
Nice, thanks for taking in count !

agaldemas:
Hi again @chicco785, just to remind you, we met at last FIWARE summit in Malaga, I came to discuss with you at the end of your QL presentation, you should have my business card...

c0c0n3:
fixed by PR #136

Issue #129, opened 2018-12-27
automatic casting to NGSI data type


chicco785:
when a value does not match the NGSI data type it should be casted it, only if the cast fails, error should be returned.

the implementation of this functionality is to be evaluated, an alternative solution could be in fact to give priority to the datatype of the value and "cast" the NGSI type to it. the two solutions are obviously incompatible.

chicco785:
see also discussion in #123

c0c0n3:
closed by #142

Issue #127, opened 2018-12-27
error : PostalAddress is not a supported NGSI Type !


agaldemas:
we have an entity using an address attribute of type PostalAddress such as:
    {
        "id": "truc_001",
        "type": "chose",
        **"address": {
            "type": "PostalAddress",
            "value": {
                "streetAddress": "18 Avenue Félix Faure",
                "postalCode": "06000",
                "addressLocality": "Nice",
                "addressCountry": "France"
            },**
            "metadata": {}
        },
        "description": {
            "type": "Text",
            "value": "test description",
            "metadata": {}
        },
        "humidity": {
            "type": "Number",
            "value": 88,
            "metadata": {}
        },
        "location": {
            "type": "StructuredValue",
            "value": {
                "type": "Point",
                "coordinates": [
                    2.185747,
                    41.3762726
                ]
            },
            "metadata": {}
        },
        "status": {
            "type": "Text",
            "value": "OK",
            "metadata": {}
        },
        "temperature": {
            "type": "Number",
            "value": 7.22,
            "metadata": {}
        }
    },

When the entity is pushed to QuantumLeap an error occur:

```
quantumleap_1  | WARNING:translators.crate:'PostalAddress' is not a supported NGSI type. Please use any of the following: Array, Boolean, ISO8601, DateTime, Integer, geo:json, geo:point, Number, Text, StructuredValue. Falling back to Text.
quantumleap_1  | [2018-12-27 15:10:59,217] ERROR in app: Exception on /v2/notify [POST]
quantumleap_1  | Traceback (most recent call last):
quantumleap_1  |   File "/usr/local/lib/python3.6/site-packages/flask/app.py", line 1982, in wsgi_app
quantumleap_1  |     response = self.full_dispatch_request()
quantumleap_1  |   File "/usr/local/lib/python3.6/site-packages/flask/app.py", line 1614, in full_dispatch_request
quantumleap_1  |     rv = self.handle_user_exception(e)
quantumleap_1  |   File "/usr/local/lib/python3.6/site-packages/flask/app.py", line 1517, in handle_user_exception
quantumleap_1  |     reraise(exc_type, exc_value, tb)
quantumleap_1  |   File "/usr/local/lib/python3.6/site-packages/flask/_compat.py", line 33, in reraise
quantumleap_1  |     raise value
quantumleap_1  |   File "/usr/local/lib/python3.6/site-packages/flask/app.py", line 1612, in full_dispatch_request
[...]

PR #132, opened 2019-01-11
Update crate.py

agaldemas:
to fix issue with PostalAddress #127 and #129
if attribute is a complex object of unknown type cast it to NGSI_STRUCTURED_VALUE, to store it as object in Crate
not finish, still miss the test of complex attribute, welcome hints on this and metadata handling

### DIFF
--- a/src/translators/crate.py
+++ b/src/translators/crate.py
@@ -234,14 +234,18 @@ class CrateTranslator(base_translator.BaseTranslator):
                 original_attrs[col] = (attr, attr_t)
 
                 if attr_t not in NGSI_TO_CRATE:
-                    supported_types = ', '.join(NGSI_TO_CRATE.keys())
-                    msg = ("'{}' is not a supported NGSI type. "
-                           "Please use any of the following: {}. "
-                           "Falling back to {}.")
-                    self.logger.warning(msg.format(
-                        attr_t, supported_types, NGSI_TEXT))
+                    # if attribute is complex assume it as 
+                    if attr:
+                        table[col] = NGSI_TO_CRATE[NGSI_STRUCTURED_VALUE]
+                    else:
+                        supported_types = ', '.join(NGSI_TO_CRATE.keys())
+                        msg = ("'{}' is not a supported NGSI type. "
+                               "Please use any of the following: {}. "
+                               "Falling back to {}.")
+                        self.logger.warning(msg.format(
+                            attr_t, supported_types, NGSI_TEXT))
 
-                    table[col] = NGSI_TO_CRATE[NGSI_TEXT]
+                        table[col] = NGSI_TO_CRATE[NGSI_TEXT]
 
                 else:
                     # Github issue 44: Disable indexing for long string


agaldemas:
Hello here is the pull request, waiting for your advice for the complex attribute test

src/translators/crate.py
```diff
@@ -234,14 +234,18 @@ def _insert_entities_of_type(self,
...
-                    supported_types = ', '.join(NGSI_TO_CRATE.keys())
-                    msg = ("'{}' is not a supported NGSI type. "
-                           "Please use any of the following: {}. "
-                           "Falling back to {}.")
-                    self.logger.warning(msg.format(
-                        attr_t, supported_types, NGSI_TEXT))
-
-                    table[col] = NGSI_TO_CRATE[NGSI_TEXT]
+                    # if attribute is complex assume it as 
+                    if attr:
```
chicco785: you should check that the value is an object, it would say

taliaga: Yes, this will make everything fall here because 'attr' is the name of the attribute, always non-empty.

I suggest this approach:

1) develop a new function that receives a python object and by python introspection returns the corresponding NGSI type constant. Add a unit-tests for it in `test_crate.py`. E.g
```
assert translator.get_ngsi_type(1) == translator.NGSI_NUMBER
assert translator.get_ngsi_type("hello") == translator.NGSI_TEXT
assert translator.get_ngsi_type({'zip': 5000, 'country': 'FOO'}) == translator.NGSI_STRUCTURED_VALUE
... (and so on for all NGSI types)
```

2) Use this function within this `if attr_t not in NGSI_TO_CRATE:` block. You'll pass the attribute's value (typically `e[attr]['value']` but it could be `e[attr]` if the entity comes in key-value pairs. (see line 227). Key-value pairs is not yet fully supported, but just to be cautious for when it will.
You may still want to use the log, just instead of falling back to `NGSI_TEXT`, you say your best-effort guess determined to be NGSI_X (the return of your function). At least during dev this will be useful. Then translate such NGSI_X and that's the crate type you'll set to table[col] like ex line 244 did.

3) The second test you'll want to add is very similar to https://github.com/smartsdk/ngsi-timeseries-api/blob/7b3902df42aaffec65cbdc2498ddf97e31128228/src/translators/tests/test_crate.py#L183, but your 'foo' will be of `PostalAddress` type and its value a complete dict, like the one your breaking use case has.

4) Make sure all translator tests run locally. I may need to improve 'contributors' documentation on how to run these, ping me when you reach this point. When translator tests pass, then check all tests (from the root of the project).

The "metadata" part mentioned in the issue may be already taken care of for you by line 234. We'll see depending on your results of step 3.

agaldemas: OK got it, I think I can go forward with your detailed advice
Thanks a lot for your reactivity and helpful hints.
I'll come to you in case of need.

chicco785:
@agaldemas, fyi failure in the test is not related to your code, @taliaga what should we do about this painful geolocation issues?

agaldemas:
I should work on the PR, during this week.

src/translators/crate.py
agaldemas: think I have managed to fix the the acceptation of unknown ngsi structured type:

function `_attr_is_structured` (issue #129)
I have added a test function: `test_accept_unknown_ngsi_type`, but not tested it locally

In addition I have fixed the issue #128  "-" in attributes and type not accepted by crate (quoting of tables and columns names, fix the problem)
I assume it's a good practice to add a test fonction then I added: `test_accept_special_chars`

I have tested on my own configuration with Orion latest, and crate 3.0.5, and it works fine 

I ran `sh run_tests.sh` from repository root directory....
Think I missed something to setup the test environment...
Time to ping you for a bit of help on this.

agaldemas:
OK, guys, I managed to fix the last remaining issue (test_capitals)
only remains your failed `test_entity_add_city_shape`
note that I had to remove lower on attributes names/columns
#128 and #129 are OK now
ready for merge

agaldemas:
Rectification, I will restore the lower case for column names, it will be better for existing database, to avoid regression ;O)

c0c0n3:
Hi guys, I've just realised that #136 overlaps with some of the work done here---namely quoting of identifiers. Merging is going to be a bit painful, but @taliaga and I will try to get it done. @agaldemas thank you so much for your contributions!!!

agaldemas:
You're welcome !

c0c0n3:
Merged into an integration branch so to be able to cherry pick only the fixes to #127 and #129 while discarding those to #128. Also moving the discussion over to #142.

c0c0n3:
follow-up: see PR #142.

PR #144, opened 2019-01-28
Use most accurate OSM result

### LINKED ISSUE #105: OSM geocoder may not return expected osm_type
Since 19 Nov 2018, the following geocoding tests fail:

* `test_entity_add_point`: OSM used to return an `osm_type` of `node` for the address used in this test but it now returns an `osm_type` of `way` which results in our `geocoding` module not being able to extract the geo-location coords from the response since we expect `node` for an address in the format `<street name> <street number> , <city> , <country code>`.
* `test_caching`: this fails because we use the same address as in `test_entity_add_point` so we can't extract the location from the OSM response and hence it won't be in the cache.

It could be that the data in OSM for that location got corrupted or a new version of Nominatim (the geocoder) went live and the API behaviour changed. Or it could be something else. But note the other full blown address we use in our tests actually works fine, i.e. we get back an `osm_type` of `node` in the response so we manage to extract the location in this case.

On another note, [it doesn't look like the OSM project is in a good shape](https://blog.emacsen.net/blog/2018/02/16/osm-is-in-trouble/)!

We have to decide how to fix this...

c0c0n3:
This PR fixes #105 by selecting the most accurate result returned by the OSM geocoder.

### DIFF
--- a/src/geocoding/tests/test_geocoding.py
+++ b/src/geocoding/tests/test_geocoding.py
@@ -9,6 +9,15 @@ import geojson
 import pytest
 
 
+def assert_lon_lat(entity, expected_lon, expected_lat):
+    assert 'location' in entity
+    assert entity['location']['type'] == 'geo:point'
+
+    lon, lat = entity['location']['value'].split(',')
+    assert float(lon) == pytest.approx(expected_lon, abs=1e-2)
+    assert float(lat) == pytest.approx(expected_lat, abs=1e-2)
+
+
 def test_entity_with_location(air_quality_observed):
     # Adding location to an entity with location does nothing
     assert 'location' in air_quality_observed
@@ -43,12 +52,7 @@ def test_entity_add_point(air_quality_observed):
     r = geocoding.add_location(air_quality_observed)
     assert r is air_quality_observed
 
-    assert 'location' in r
-    assert r['location']['type'] == 'geo:point'
-
-    lon, lat = r['location']['value'].split(',')
-    assert float(lon) == pytest.approx(51.2358357, abs=1e-2)
-    assert float(lat) == pytest.approx(4.4201911, abs=1e-2)
+    assert_lon_lat(r, expected_lon=51.23, expected_lat=4.42)
 
 
 def test_entity_add_point_negative_coord(air_quality_observed):
@@ -64,12 +68,7 @@ def test_entity_add_point_negative_coord(air_quality_observed):
     r = geocoding.add_location(air_quality_observed)
     assert r is air_quality_observed
 
-    assert 'location' in r
-    assert r['location']['type'] == 'geo:point'
-
-    lon, lat = r['location']['value'].split(',')
-    assert float(lon) == pytest.approx(19.5115112, abs=1e-2)
-    assert float(lat) == pytest.approx(-99.0883494, abs=1e-2)
+    assert_lon_lat(r, expected_lon=19.51, expected_lat=-99.08)
 
 
 def test_entity_add_street_line(air_quality_observed):
@@ -127,7 +126,8 @@ def test_entity_add_country_shape(air_quality_observed):
 
     geo = r['location']['value']
     assert geo['type'] == 'MultiPolygon'
-    assert len(geo['coordinates']) == pytest.approx(12, abs=2)
+    multi_polygon = geojson.MultiPolygon(geo['coordinates'])
+    assert multi_polygon.is_valid
 
 
 def test_multiple_entities(air_quality_observed):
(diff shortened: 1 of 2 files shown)


chicco785:
LGTM

src/geocoding/tests/test_geocoding.py
```diff
@@ -108,8 +108,8 @@ def test_entity_add_city_shape(air_quality_observed):
 
     geo = r['location']['value']
     assert geo['type'] == 'Polygon'
-    assert len(geo['coordinates']) == 1
-    assert len(geo['coordinates'][0]) == pytest.approx(2186, abs=100)
+    polygon = geojson.Polygon(geo['coordinates'])
+    assert polygon.is_valid
```
chicco785: so we now just check that the polygon is indeed a polygon regardless what polygon and what polygon, right?

src/geocoding/tests/test_geocoding.py
```diff
@@ -68,8 +68,8 @@ def test_entity_add_point_negative_coord(air_quality_observed):
     assert r['location']['type'] == 'geo:point'
 
     lon, lat = r['location']['value'].split(',')
-    assert float(lon) == pytest.approx(19.6389474, abs=1e-2)
-    assert float(lat) == pytest.approx(-98.9109537, abs=1e-2)
+    assert float(lon) == pytest.approx(19.5115112, abs=1e-2)
```
taliaga: was this one failing or this is just an adjustment of precision? If the former, we should go for the "at least one of many is correct" approach

chicco785: do we actually need to test this? or better should the test fail? IMHO, this is like "testing" the geo-decoding service being correct, which is out of our scope (I would say).

taliaga: we don't need to test the exact value used by OSM, but we need to test we couple well with whatever response they give, at least "in most cases" (hence my comment above).

That's why we don't need much precision in the test, just enough to recognise we're at least not mixing lat and long, things like that.

chicco785: understood and agreed, so purpose being to test if we support correctly returned: point, line, polygon, right? but still, if the change something for a search, we will end up in trouble. can we do a mockup test?

chicco785: i am thinking how to avoid that a change in the data of geocoding service screw up our integration tests

taliaga: > can we do a mockup test?

unfortunately mocking the response would not work because a "structural change" from their side would not make our test fail.

> i am thinking how to avoid that a change in the data of geocoding service screw up our integration tests

This is the key question. My suggestion discussed with @c0c0n3 was to choose say 5 points of big reliable cities and pass the test as long as at least one of them is correct (with gross tolerance, not being picky there). 

I didn't quite get yet why "picking the most accurate result" would solve the real issue (hence my question to @c0c0n3) :)

c0c0n3: yep, I fully agree with you guys on this: we should test structure rather than semantics---see my comment above. For example, I fixed the other test just by checking that what we get back is a polygon, rather than checking it is the polygon enclosing Mexico City. The reason I didn't do it for this too is that I'd like to check if my theory is correct. Let me explain what I have in mind.

If you look at the changes I made in the geocoding module, you'll see we're now picking the best match in terms of coordinates accuracy. I think these tests were breaking because of OSM returning *every time* the same result set for our geo query but with entries in a *different* order as in

    1. where is P? It's at: [(X, with accuracy ~ 0.1), (Y, with accuracy ~ 0.2)]
    2. where is P? It's at: [(Y, with accuracy ~ 0.2), (X, with accuracy ~ 0.1)]

If I'm right about this, then this test shouldn't fail going forward. If it fails again, it's probably a case of data corruption on the OSM side.

In any case, going forward I suggest we do the following:

1. Define what is the interface between us and a map provider---the `geocoder` lib isn't quite clear about what such an interface should be in terms of data being exchanged.
2. Provide a component implementing that interface by sourcing data from OSM.
3. Make the rest of our code work against that interface.
4. Use a mocked version of that interface for our unit tests.
5. Write *separate* integration tests to verify OSM works as (2) expects---this would catch OSM breaking API changes.

We basically have 2 and 3 in place. We need a bit more work for the rest which is why I haven't done it as part of this PR. Also, I've been thinking about test case generation using an approach similar to this: https://www.well-typed.com/blog/2019/01/qsm-in-depth/

But I suggest we talk about all this next time meet.

c0c0n3: Implemented relaxed coords checks as discussed earlier---see d98149b.

c0c0n3:
@taliaga PR should be ready to merge. Will do as soon as you give me the green light...

Issue #145, opened 2019-01-29
Error when receiving empty value for datetime types


taliaga:
When notifications come with attributes of type DateTime (or ISO8601) and no value, QL fails to insert a default "null".

For reference, see https://github.com/smartsdk/ngsi-timeseries-api/pull/122/files

pooja1pathak:
Hi @taliaga @c0c0n3 ,
Is this issue fixed by PR https://github.com/smartsdk/ngsi-timeseries-api/pull/150 ? or need to be fixed?

c0c0n3:
@pooja1pathak yes good catch, PR #150 fixes it, we just plain forgot to close this issue. Thanks!
closed by #150.

PR #146, opened 2019-01-29
Fix GH #145

taliaga:


src/reporter/reporter.py
```diff
@@ -60,8 +60,16 @@ def has_value(entity, attr_name):
     attr_value = attr.get('value', None)
     attr_type = attr.get('type', None)
 
+    if attr_value is None:
+        return False
+
+    # GH Issue #145
+    if attr_type in (NGSI_DATETIME, NGSI_ISO8601):
+        return attr_value.strip() != ''
```
chicco785: i am unsure what this actually does :) is the bug about the space among ' ' ?

taliaga: That's right, the spaced empty (typically seen in payloads from IoT Agents) was not being recognised as Null and QL attempted to translate it into a timestamp.

There's another github issue aimed to solve this more robustly (like in: what if the datetime string is "foo"). But my understanding is that this was needed asap.

chicco785: indeed

PR #158, opened 2019-02-01
Fix security vulnerability CVE-2017-18342.

### LINKED ISSUE #131: fix CVE-2017-18342
CVE-2017-18342 [More information](https://nvd.nist.gov/vuln/detail/CVE-2017-18342)
high severity
Vulnerable versions: < 4.2b1
Patched version: 4.2b1
In PyYAML before 4.1, the yaml.load() API could execute arbitrary code. In other words, yaml.safe_load is not used.

c0c0n3:
This PR fixes #131.

Specifically there's a security vulnerability in `PyYAML` that could allow an attacker to execute arbitrary code using a YAML file. Even though we don't use `PyYAML` directly, `connextion`, one of our deps, does as does `clickclick`, a lib `connexion` depends on.

At the moment, there's no official `PyYAML` release with the fix, only pre-release versions. It seems `PyYAML` hasn't been getting much love in the last couple of years, but fortunately they're [planning to release version 4.2][pyyaml-4.2] which should include the fix for CVE-2017-18342.

Since there's no official `PyYAML 4.2`, the best we can do for the time being is to bring in a patched version through an explicit dependency in our `Pipfile`:

    pyyaml = ">=4.2b1"

But this has the downside we now have to generate the lock file with

    $ pipenv lock --pre

which tells `pipenv` to consider pre-release/patched versions too when resolving dependencies. So the next time we generate the lock file we might end up with some potentially unstable or not-yet-fully-polished libs in our dependency chain. (This is **not** the case at the moment.)

The good news is that as soon as `PyYAML 4.2` comes out, we can remove the explicit dependency and generate the lock file as usual, **without** the `--pre` flag. In fact, both `connexion` and `clickclick` specify a dependency on `PyYAML > 3` so that if `4.2` is available in PyPi then `pipenv` will use it.

Also note I had to add another explicit dependency on a library we don't use directly:

    certifi = "==2018.10.15"

it's got to be pinned otherwise `pipenv` will use a later version which doesn't work with `requests`. For the record, this is the error I get:

    ImportError while loading conftest '/src/ngsi-timeseries-api/src/conftest.py'.
     conftest.py:6: in <module>
         import requests
    /usr/local/lib/python3.6/site-packages/requests/__init__.py:112: in <module>
         from . import utils
    /usr/local/lib/python3.6/site-packages/requests/utils.py:24: in <module>
         from . import certs
    /usr/local/lib/python3.6/site-packages/requests/certs.py:15: in <module>
         from certifi import where
    E   ModuleNotFoundError: No module named 'certifi'

[pyyaml-4.2]: https://github.com/yaml/pyyaml/issues/193

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -19,6 +19,12 @@ numpy = "~=1.12"
 pymongo = "~=3.4"
 pytest-flask = "~=0.10"
 
+# Indirect dependencies needing special handling.
+# See #157 for the details.
+# Note. Run `pipenv lock --pre` to generate the lock file.
+pyyaml = ">=4.2b1"
+certifi = "==2018.10.15"
+
 [dev-packages]
 
 [requires]
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "2c7fbe11a509609ae3237766fba2bd9395b3ac171634b21c789c53360af12445"
+            "sha256": "e1dc8073d2733b555226f40268a8b339af6c77e9ad89cfbf0824a85b3da48598"
         },
         "pipfile-spec": 6,
         "requires": {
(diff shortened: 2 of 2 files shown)


c0c0n3:
closes #131.

Issue #93, opened 2018-10-19
Update time_index generation policy


taliaga:
As regards method `_get_time_index`,

With higher priority than `dateModified` it should look for the `timestamp` in the metadata (https://github.com/Fiware/dataModels/pull/302/files#diff-9308877b8fc4a861aa6d718690d52f82R331) and the `TimeInstant` element, in the case of https://github.com/telefonicaid/iotagent-node-lib#the-timeinstant-element

agaldemas:
Hello,
We need this feature to be able to store device timestamped data, coming from IOT agents, through `TimeInstant` element, 
may I/you open a new PR, where I can fix this behavior: Look for & give priority to `TimeInstant` over `dateModified` could be a good title ?
Nevertheless by experience it's better to keep any timestamp information we can grab for analytics, so I suggest to keep and store `dateModified` from Orion notification

chicco785:
Hi,
This is something we have been discussing: how to specify a different attribute metadata as TimeIndex. We have been thinking of adding some metadata to the notification to handle that.

A PR is welcome, of course. Let’s open a google doc to collaborative design the solution, what do you think?

Cheers,
Federico

agaldemas:
Sorry, for late reply, I went through other issues with orion ngsi behavior regarding milliseconds in timestamp (DateTime type)...
I wanted to be sure that we could provide an efficient way to retrieve milliseconds from Orion notification, given an update context sent by an iot agent, and TimeInstant attribute, 
since the aim is to be able to handle device attributes timestamp at milliseconds/microseconds resolution.
  After some testing, I realized there is no way for the moment except using metadata, confirmed by @fgalan  in an issue discussion on the subject: [support for milliseconds..](https://github.com/telefonicaid/fiware-orion/issues/3412).
Think we should invite him in the discussion ?
What do you think ?

chicco785:
@c0c0n3 is landing shortly a pr to configure using headers the attribute to be used as timeindex. (Which means TimeInstant will be supported). This could be a starting point, but it will not cover metadata (simply because as of today, ql does not store metadata). TimeIndex is (if I recall correctly) stored as epoch timestamp. Supporting microseconds may be not possible, it depends on what crate.io allows.

taliaga:
That's right, the database stores time_index as a `long` representing milliseconds since epoch.
[...]

PR #161, opened 2019-02-11
Time index selection

c0c0n3:
This PR implements a strategy to select a time index value from the notification payload and implements the features requested in #93.

### Selection strategy
We determine which attribute or metadata value to use as a time index for the entity being notified by selecting the first value found in the below list that can be converted to a `datetime`. Items are considered from top to bottom, so that if multiple values are present and they can all be converted to `datetime`, the topmost value is chosen.

1. Custom time index. The value of the `Fiware-TimeIndex-Attribute`. Note that for a notification to contain such header, the corresponding subscription has to be created with an `httpCustom` block as detailed in the *Subscriptions* and *Custom Notifications* sections of the [NGSI spec][ngsi-spec].
2. `TimeInstant` attribute. As specified in the Fiware [IoT agent documentation][time-instant].
3. `TimeInstant` metadata. The most recent `TimeInstant` attribute value found in any of the attribute metadata sections in the notification. (Again, refer to the [IoT agent documentation][time-instant].)
4. `dateModified` attribute.
5. `dateModified` metadata. The most recent `dateModified` attribute value found in any of the attribute metadata sections in the notification.
6. Current time. This is the default value we use if any of the above isn't present or none of the values found can actually be converted to a `datetime`.

### API extensions
Since *QuantumLeap* has a `subscribe` endpoint to set up NGSI subscriptions, we have extended it to support setting up a subscription where you can easily specify the time index to use in notifications: the time index attribute to use is simply passed in as an additional query parameter to the existing  `subscribe` endpoint and with that *QantumLeap* creates a custom NGSI subscription containing the `Fiware-TimeIndex-Attribute` as detailed earlier.

### Time representation considerations
Both `TimeInstant` and `dateModified` are supposed to contain timestamps represented in the ISO 8601 format. (See: [IoT agent documentation][time-instant] and `dateModified ` in the *Built-in Attributes* section of the [NGSI spec][ngsi-spec].) We have introduced timestamp validation so that *QantumLeap* will use a timestamp value as time index *only* if it actually is a valid ISO 8601 timestamp. This makes one of our regression tests (`test_index_with_millis`) fail since it uses notifications containing `dateModified` attributes whose values are expressed as milliseconds from the epoc as opposed to ISO 8601 timestamps. We have to decide if we should support this too or rather stick to the NGSI spec which only caters for ISO 8601 timestamps.

[ngsi-spec]: http://fiware.github.io/specifications/ngsiv2/stable/
[time-instant]: https://github.com/telefonicaid/iotagent-node-lib#the-timeinstant-element

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -6,6 +6,7 @@ name = "pypi"
 [packages]
 connexion = "~=1.1"
 crate = "~=0.22"
+python-dateutil = ">=2.7"
 flask = "~=0.12"
 geocoder = "~=1.33"
 geojson = "~=2.4"
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "e1dc8073d2733b555226f40268a8b339af6c77e9ad89cfbf0824a85b3da48598"
+            "sha256": "4a33a5f38735c73186f90433ab72ae2a59998b33e8e82c7626a334c8eab2832e"
         },
         "pipfile-spec": 6,
         "requires": {
@@ -376,10 +376,11 @@
         },
         "python-dateutil": {
             "hashes": [
-                "sha256:063df5763652e21de43de7d9e00ccf239f953a832941e37be541614732cdfc93",
-                "sha256:88f9287c0174266bb0d8cedd395cfba9c58e87e5ad86b2ce58859bc11be3cf02"
+                "sha256:7e6584c74aeed623791615e26efd690f29817a27c73085b78e4bad02493df2fb",
+                "sha256:c89805f6f4d64db21ed966fda138f8a5ed7a4fdbc1a8ee329ce1b74e3c74da9e"
             ],
-            "version": "==2.7.5"
+            "index": "pypi",
+            "version": "==2.8.0"
         },
         "pytz": {
             "hashes": [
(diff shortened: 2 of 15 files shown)


c0c0n3:
@taliaga as promised here are a couple of reminders about the stuff we discussed earlier:

#### timestamps
If it turns out we should support timestamps as millis from the epoc, we can easily do that by adding a couple of lines to the `to_datetime` method. (See e.g. [this](https://stackoverflow.com/questions/748491) and [this](https://stackoverflow.com/questions/6999726).)

#### tests
It's probably a good idea to add a couple more integration tests. Specifically, one test should check that after setting up a subscription in Orion with a custom header (i.e. the time index attribute), we actually get that header in any subsequent notification. The PR already comes with a test to check we can use custom subscriptions with Orion and one to check that if we get notified with a time index custom header we actually use it. But it's a good thing to go the extra mile and add this test too I suppose. And as we're at it, we might just as well check if Orion will ever pass on `TimeInstant` values an IoT agent originally put in the metadata.

#### documentation
It's probably a good idea to document the time index selection strategy at the API level too---perhaps in the Swagger file. Once you've found a good home for it, then I could just apply my famed C&P technology to copy the info from the **selection strategy** section above in this PR.

chicco785:
i am unsure why the builds fails

taliaga:
As discussed in person with Andrea, we will keep supporting time_index only in ISO format, as defined in the API spec. The test was checking support for millis since epoch, though that was not in the spec and it's prone to timezones issues. It's not impossible to support such input if required in the future. We'll keep it simple for now.

taliaga:
This PR looks good to me as it is. I'll add more tests (the ones we discussed about, like what happens if the specified attr in the header is not in the notification payload, etc) and extend the user-docs better explaining this new strategy in a following PR, probably the one preparing the 0.6 release.

PR #163, opened 2019-02-13
Fix _adjust_gh_44 issue

taliaga:


chicco785:
it looks to me this fix is already in v0.6 so i am unsure why a new pr is needed.

taliaga:
That's right. The thing is important (hot) fixes for master should ideally not wait until next releases of new features. Since I had already this branch (and its tag) being used, I preferred merging rather than dropping it.

PR #176, opened 2019-03-19
Time_index enhancement (issue 172)

heikkilv:
Updated <code>time_index</code> selection with three new steps (2, 5 and 6):

### Selection strategy

We determine which attribute or metadata value to use as a time index for the entity being notified by selecting the first value found in the below list that can be converted to a datetime. Items are considered from top to bottom, so that if multiple values are present and they can all be converted to datetime, the topmost value is chosen.

1. Custom time index. The value of the <code>Fiware-TimeIndex-Attribute</code>. Note that for a notification to contain such header, the corresponding subscription has to be created with an httpCustom block as detailed in the Subscriptions and Custom Notifications sections of the [NGSI spec](http://fiware.github.io/specifications/ngsiv2/stable/).
2. Custom time index metadata. The most recent custom time index (the value of the <code>Fiware-TimeIndex-Attribute</code>) attribute value found in any of the attribute metadata sections in the notification.
3. <code>TimeInstant</code> attribute. As specified in the Fiware [IoT agent documentation](https://github.com/telefonicaid/iotagent-node-lib#the-timeinstant-element).
4. <code>TimeInstant</code> metadata. The most recent <code>TimeInstant</code> attribute value found in any of the attribute metadata sections in the notification. (Again, refer to the [IoT agent documentation](https://github.com/telefonicaid/iotagent-node-lib#the-timeinstant-element).)
5. <code>timestamp</code> attribute.
6. <code>timestamp</code> metadata. The most recent <code>timestamp</code> attribute value found in any of the attribute metadata sections in the notification. As specified in the Fiware [data models documentation](https://fiware-datamodels.readthedocs.io/en/latest/guidelines/index.html#dynamic-attributes).
7. <code>dateModified</code> attribute.
8. <code>dateModified</code> metadata. The most recent <code>dateModified</code> attribute value found in any of the attribute metadata sections in the notification.
9. Current time. This is the default value we use if any of the above isn't present or none of the values found can actually be converted to a datetime.

### DIFF
--- a/src/reporter/subscription_builder.py
+++ b/src/reporter/subscription_builder.py
@@ -7,6 +7,10 @@ def build_subscription(quantumleap_url,
                        attributes, observed_attributes, notified_attributes,
                        throttling_secs,
                        time_index_attribute=None):
+    metadata_list = ['dateCreated', 'dateModified', 'TimeInstant', 'timestamp']
+    if time_index_attribute is not None:
+        metadata_list.append(time_index_attribute)
+
     tree = subscription(
         description('Created by QuantumLeap {}.'.format(quantumleap_url)),
         subject(
@@ -22,7 +26,7 @@ def build_subscription(quantumleap_url,
         ),
         notification(
             build_notification_target(quantumleap_url, time_index_attribute),
-            metadata(['dateCreated', 'dateModified', 'TimeInstant']),
+            metadata(metadata_list),
             attrs(first_of(attributes, notified_attributes))
         ),
         throttling(throttling_secs)
--- a/src/reporter/tests/test_subscribe.py
+++ b/src/reporter/tests/test_subscribe.py
@@ -43,7 +43,7 @@ def test_valid_defaults(clean_mongo, clean_crate):
         'attrs': [],
         'attrsFormat': 'normalized',
         'http': {'url': "{}/notify".format(QL_URL)},
-        'metadata': ['dateCreated', 'dateModified', 'TimeInstant'],
+        'metadata': ['dateCreated', 'dateModified', 'TimeInstant', 'timestamp']
     }
     assert subscription['throttling'] == 1
 
@@ -82,7 +82,7 @@ def test_valid_customs(clean_mongo, clean_crate):
         'attrs': ["pressure"],
         'attrsFormat': 'normalized',
         'http': {'url': "{}/notify".format(QL_URL)},
-        'metadata': ['dateCreated', 'dateModified', 'TimeInstant'],
+        'metadata': ['dateCreated', 'dateModified', 'TimeInstant', 'timestamp']
     }
     assert subscription['throttling'] == 30
 
@@ -155,5 +155,5 @@ def test_custom_time_index(clean_mongo, clean_crate):
                 TIME_INDEX_HEADER_NAME: 'my-time-index'
             }
         },
-        'metadata': ['dateCreated', 'dateModified', 'TimeInstant'],
+        'metadata': ['dateCreated', 'dateModified', 'TimeInstant', 'timestamp', 'my-time-index']
     }
--- a/src/reporter/tests/test_timex.py
+++ b/src/reporter/tests/test_timex.py
@@ -2,7 +2,9 @@ from datetime import *
 from reporter.timex import *
 
 
-def build_notification(custom, ti, dm, mti, mdm, a1_ti, a1_dm, a2_ti, a2_dm):
+def build_notification(custom, ti, ts, dm,
+                       a1_custom, a1_ti, a1_ts, a1_dm,
+                       a2_custom, a2_ti, a2_ts, a2_dm):
     return {
         'customTimeIndex': {
             'value': custom
@@ -10,21 +12,22 @@ def build_notification(custom, ti, dm, mti, mdm, a1_ti, a1_dm, a2_ti, a2_dm):
         'TimeInstant': {
             'value': ti
         },
+        'timestamp': {
+            'value': ts
+        },
         'dateModified': {
             'value': dm
         },
-        'metadata': {
-            'TimeInstant': {
-                'value': mti
-            },
-            'dateModified': {
-                'value': mdm
-            }
-        },
         'a1': {
             'metadata': {
+                'customTimeIndex': {
+                    'value': a1_custom
+                },
                 'TimeInstant': {
                     'value': a1_ti
+                },
+                'timestamp': {
+                    'value': a1_ts
                 },
                 'dateModified': {
                     'value': a1_dm
@@ -33,8 +36,14 @@ def build_notification(custom, ti, dm, mti, mdm, a1_ti, a1_dm, a2_ti, a2_dm):
         },
         'a2': {
             'metadata': {
+                'customTimeIndex': {
+                    'value': a2_custom
+                },
                 'TimeInstant': {
                     'value': a2_ti
+                },
+                'timestamp': {
+                    'value': a2_ts
                 },
                 'dateModified': {
                     'value': a2_dm
@@ -46,7 +55,7 @@ def build_notification(custom, ti, dm, mti, mdm, a1_ti, a1_dm, a2_ti, a2_dm):
 
 def build_notification_timepoints(base_point):
     ts = []
-    for k in range(9):
+    for k in range(12):
         d = base_point + timedelta(days=k)
         ts.append(d.isoformat())
     return ts
@@ -64,6 +73,22 @@ def test_custom_index_takes_priority():
         select_time_index_value(headers, notification)
 
 
+def test_use_latest_meta_custom_index():
+    headers = {
+        TIME_INDEX_HEADER_NAME: 'customTimeIndex'
+    }
+    base_point = datetime(2019, 1, 1)
+
+    ts = build_notification_timepoints(base_point)
+    ts[0] = None  # custom index attribute slot
+
+    notification = build_notification(*ts)
+
+    # latest meta custom index slot = 8
+    assert ts[8] == \
+        select_time_index_value(headers, notification).isoformat()
+
+
 def test_skip_custom_index_if_it_has_no_value():
     headers = {
         TIME_INDEX_HEADER_NAME: 'customTimeIndex'
@@ -71,6 +96,8 @@ def test_skip_custom_index_if_it_has_no_value():
     base_point = datetime(2019, 1, 1)
     ts = build_notification_timepoints(base_point)
     ts[0] = None  # custom index slot
+    ts[4] = None  # custom index metadata for a1
+    ts[8] = None  # custom index metadata for a2
     notification = build_notification(*ts)
 
     assert ts[1] == \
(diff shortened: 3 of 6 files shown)


chicco785:
@taliaga all probably we should measure the latency introduced by applying the strategy.

Feelings? (Not much for this addition, but in general)

taliaga:
Thank you @heikkilv, nice work!. I'll just drop some minor comments on the diffs. @c0c0n3 should too!.

> probably we should measure the latency introduced by applying the strategy. 

I don't believe this PR is adding "latency" to the strategy as it is now, but certainly we could measure how it changed from the days we only used `dateModified`. Unfortunately we don't have performance tests yet, but we could add some as a way to instrument a constant profiling.

> Feelings? (Not much for this addition, but in general)

I think we could simplify a bit the way we explain this policy (given we always look at value first metadata then). But all in all it looks good enough, at least up to the use-cases I'm aware of.

When needed, to have a more "complete" coverage, we'd need some cross-attribute options to pick the time-index (like use the oldest/newest timestamp of any of the attributes, etc). So basically users would just need to pick at most 3 things:
1 - The name of the custom attribute they want to use (if any)
2 - To consider or ignore the content in "metadata" section.
3 - If he only cares of "latest" or "oldest" timestamp regardless of the attribute and whether it is value or metadata.
Combining these 3 it'll be possible to cover a wide range of cases I think.

chicco785:
in fact my comment was: "not for this addition, but in general". Point being that this way of selecting the attribute, in the "worst" case scenario, may introduce some delay, and it may better to introduce an option to define from the usecase perspective which "policy" order ql should use. just an idea... but if in all the cases the "latency" introduced is negligible, it will not be worth to add this feature.

As a todo, can't we think of a test suite that running a test for each "case" measure the time between data arrival and data population in the backend?

src/reporter/tests/test_subscribe.py
```diff
@@ -155,5 +155,5 @@ def test_custom_time_index(clean_mongo, clean_crate):
                 TIME_INDEX_HEADER_NAME: 'my-time-index'
             }
         },
-        'metadata': ['dateCreated', 'dateModified', 'TimeInstant'],
+        'metadata': ['dateCreated', 'dateModified', 'TimeInstant', 'timestamp', 'my-time-index']
```
taliaga: are these additions to the metadata part of the subscription really needed? what's the consequence of adding them? (i.e, is Orion doing anything with those?)

src/reporter/tests/test_timex.py
```diff
@@ -64,13 +73,31 @@ def test_custom_index_takes_priority():
...
+
+
 def test_skip_custom_index_if_it_has_no_value():
     headers = {
         TIME_INDEX_HEADER_NAME: 'customTimeIndex'
     }
     base_point = datetime(2019, 1, 1)
     ts = build_notification_timepoints(base_point)
     ts[0] = None  # custom index slot
+    ts[4] = None  # custom index metadata for a1
```
taliaga: How about defining have a bunch of module-level constants to represent these options and maybe a list containing all of them (or not) so we prevent confusions of indexing and duplicated comments? Something like what we see in `test_use_date_modified` but more visible.

src/reporter/timex.py
```diff
@@ -70,9 +46,14 @@ def select_time_index_value(headers: dict, notification: dict) -> datetime:
       subscription has to be created with an ``httpCustom`` block as detailed
       in the *Subscriptions* and *Custom Notifications* sections of the NGSI
       spec.
+    - Custom time inde metadata. The most recent custom time index attribute value
```
taliaga: typo: inde

src/reporter/timex.py
```diff
@@ -70,9 +46,14 @@ def select_time_index_value(headers: dict, notification: dict) -> datetime:
...
+      found in any of the attribute metadata sections in the notification.
     - ``TimeInstant`` attribute.
     - ``TimeInstant`` metadata. The most recent ``TimeInstant`` attribute value
       found in any of the attribute metadata sections in the notification.
+    - ``timestamp`` attribute.
+    - ``timestamp`` metadata. The most recent ``timestamp`` attribute value
+      found in any of the attribute metadata sections in the notification.
```
taliaga: Some pending items:
- We'd need to update the docs in https://github.com/smartsdk/ngsi-timeseries-api/blob/master/docs/manuals/user/index.md
- Target PR towards 0.6.1rc (updating release notes)

src/reporter/timex.py
```diff
@@ -84,23 +65,27 @@ def select_time_index_value(headers: dict, notification: dict) -> datetime:
...
+        attr_value = to_datetime(_attribute(notification, attr_name))
+        if attr_value:
+            return attr_value
+
+        meta_value = latest_from_str_rep(
+            _iter_metadata(notification, attr_name))
+        if meta_value:
+            return meta_value
+
+    return datetime.now()
```
taliaga: From chicco's comment, this `now()` call might be better done at the beginning of the whole method.

src/reporter/timex.py
```diff
@@ -84,23 +65,27 @@ def select_time_index_value(headers: dict, notification: dict) -> datetime:
...
-        custom_index, time_instant, meta_time_instant, date_modified,
-        meta_date_modified, default_value
+    considered_attributes = [
+        maybe_value(headers, TIME_INDEX_HEADER_NAME),
+        "TimeInstant",
+        "timestamp",
+        "dateModified"
     ]
 
-    return [d for d in priority_list if d][0]
```
c0c0n3: Um, why trade a simple algorithm and self-documenting code for the apparently more complicated loop down below? Isn't it possible to just add another entry to the priority list for "timestamp"? 
Just trying to understand the reasoning behind it, since I've most likely missed something... :-)

src/reporter/timex.py
c0c0n3: why? sorry I missed that too :-)

src/reporter/tests/test_timex.py
c0c0n3: yah, them tests are a disaster area, sorry I whipped them together way too quickly...they should be rewritten properly...

src/reporter/tests/test_subscribe.py
heikkilv: I have understood it so that only those metadata attributes that are in the subscription list are included by Orion in the notification. For example, checking for <code>timestamp</code> in metadata (as in step 6) is useless unless it is included in subscription list.

src/reporter/timex.py
heikkilv: Yes, to be more accurate the current time should really be saved at the beginning, since processing the time index calculation might take a little time. Should be only a couple of milliseconds at most but still a good point.

src/reporter/timex.py
heikkilv: I changed the priority list algorithm because there is no need to calculate all the entries for the priority list and then only return the first item that has a value. Instead, it should be more efficient to calculate the entries one by one and return immediately when a proper value is found.

I made the loop because I didn't want to copy paste the same thing 4 times. I.e. the custom attribute, <code>TimeInstant</code>, <code>timestamp</code> and <code>dateModified</code> are all handled in the same way.

src/reporter/timex.py
heikkilv: I moved the target from master to 0.6.1rc and updated the documentation with the selection process. The subscription part should probably also be updated. Right now it only mentions <code>dateCreated></code> and <code>dateModified</code> for the metadata.

By the way, is the <code>dateCreated</code> metadata actually used anywhere in QL?

src/reporter/timex.py
c0c0n3: Yah, I see your point, thanks for clarifying. But in terms of efficiency, quite frankly I don't think it's going to make a difference in practical terms. If you try running e.g.

    python -m pytest -vv --durations=0 reporter/tests/test_timex.py

you'll see the reported duration of each test is 0.00s---I bet you we're talking microseconds here, not even millis! So I'd rather keep the earlier approach since it seems cleaner? Also both approaches require linear time to get the job done---well assuming dictionary lookups are `O(1)` which they should I think. So in practice I reckon you'd need a JSON payload with hundreds of attributes to appreciate any difference in performance. On another note, a quick way to keep the old code and the performance improvement you suggest would be to make `priority_list` lazy---i.e. using generators, as in `custom_index = yield to_datetime(...)`.

Anyhoo, if you'd rather keep your code the way it is, I'm fine with that too :-)

heikkilv: I changed the algorithm to be more explicit and hopefully more readable.

c0c0n3: @heikkilv wow! awesome sauce! thank you sooo much for this, much appreciated!!!

src/reporter/tests/test_subscribe.py
taliaga: I see, let me better test and understand what's the implications of adding these things to the metadata part. In other words, if we don't add these new items, does something change? I'll check and get back to you.

src/reporter/tests/test_timex.py
taliaga: no worries! I just commented to check if @heikkilv , while at it, wouldn't mind doing this mini-change (only in this module at least) If not, we can do it later.

src/reporter/timex.py
taliaga: excellent, thanks! good question, IIRC we're not using the `dateCreated` at all, so this might be something worth pruning at some point (not now). Again, this will clearer come with a better understanding of what really happens in Orion with the elements in the 'metadata' section of the subcription (see my other comment)

src/reporter/timex.py
taliaga: lovely

src/reporter/tests/test_timex.py
heikkilv: I too think that the tests need a bigger rewrite. I don't think it is useful to modify them further with this pull request.

src/reporter/subscription_builder.py
```diff
@@ -22,7 +26,7 @@ def build_subscription(quantumleap_url,
         ),
         notification(
             build_notification_target(quantumleap_url, time_index_attribute),
-            metadata(['dateCreated', 'dateModified', 'TimeInstant']),
+            metadata(metadata_list),
```
taliaga: I believe you don't need to change this, right?. All you want to change is in time_index's (`timex.py`) side of QuantumLeap. Afaik, Orion will not interpret any new data in this `metadata` list other than `dateCreated` and `dateModified` (I still need to check what about `TimeInstant`). In other words, this change seems to have no effect to me, then we don't need to update the expected 'metadata' in the tests neither.

src/reporter/tests/test_subscribe.py
taliaga: see my comment above

src/reporter/subscription_builder.py
heikkilv: Orion only sends those metadata attributes that are in list with the notification. See [http://telefonicaid.github.io/fiware-orion/api/v2/stable/](http://telefonicaid.github.io/fiware-orion/api/v2/stable/) -> Specification -> Filtering out attributes and metadata

So if you use <code>'metadata': ['dateCreated', 'dateModified']</code> when making the subscription, <code>timex.py</code> cannot access <code>TimeInstant</code>, <code>timestamp</code> or a custom meta attribute because they are not included in the notification. In that case, it would end up using <code>dateModified</code> but that might be totally different than the other time-related meta attributes.

But when you use <code>'metadata': ['dateCreated', 'dateModified', 'TimeInstant', 'timestamp']</code> when making the subscription, <code>TimeInstant</code> and <code>timestamp</code> will be included in the notification and can be used in determining the time index. Of course, <code>TimeInstant</code> and <code>timestamp</code> will only be included by Orion if they actually exist within the entity. Any custom meta attribute would have to included similarly in the subscription.

taliaga: Cool, I didn't manage to test if `TimeInstant` and `timestamp` had special treatment by Orion. So, in that case, then it's fine to keep them there. I just tested the inclusion of the custom attribute in metadata list of the subscription and noticed that didn't make Orion include it at all. What QL will have to do is to include the custom attribute instead of in the metadata list, in the notification's attrs list. The user should do that of course, but QL's doing it is a safer play since the user already specified it in the header. Does it make sense?

heikkilv: AFAIK the <code>dateCreated</code> and <code>dateModified</code> are the only attributes that have special treatment in Orion. I.e. they are automatically created as entity level attributes as well as attribute level meta attributes.

To make use of the step 1 of the selection strategy (i.e. custom attribute), the custom attribute must be included in the notication's attrs list as you said. The same is with the <code>TimeInstant</code> attribute (step 3), the <code>timestamp</code> attribute (step 5) and the <code>dateModified</code> attribute (step 7). Unless they are included in the attrs list, the notification will not include those attributes and obviously those steps in the selection strategy cannot be used. This is similarly to the steps concerning the meta attributes (steps 2, 4, 6 and 8) that need the attribute names in notification metadata list.

Note that empty list (or missing list) means that all attributes (or all metadata) are included in the notification with the exception of the built-in attributes <code>dateCreated</code> and <code>dateModified</code>. To include those they need to be specifically included by the lists as is explained in the link in my previous comment.

Instead of including all those attributes in the notification attrs list, I would rather leave it up to the user. That way the user can have more control on the time index selection. By default i.e. with parameters <code>attributes=None</code> and <code>notified_attributes=None</code> for the subscription builder, the attribute list is omitted, meaning that all attributes will be included in the notification. That of course includes the user selected custom attribute as well.

taliaga: A cause of confusion was the dilemma of whether including an attr in the metadata list of the subscription would add such attribute in the notification or not. We conclude it will add it in the notification but in the metadata section only (meta attribute) provided it existed, not as entity-level attributes (this is just managed by attrs list of the subscription). IIRC, the case of `dateModified` / `dateCreated` is just slightly different because even with an empty list in the attrs list of the subscription (i.e., asking it all), they still are needed in the metadata list if someone wants them to appear as "entity level attributes".

You're right, with default values for the subscription ensuring the custom attribute sent via header will arrive in the notification at entity-level (and metadata level) we're good.

Thanks again for your PR and patience in the discussion :)

PR #178, opened 2019-04-16
Fix GH#177

### LINKED ISSUE #177: Bug: throttling cannot be set to zero
Using the subscribe endpoint as described in https://github.com/smartsdk/ngsi-timeseries-api/blob/master/specification/quantumleap.yml#494, if the throttling query parameter is set to zero, QL creates the subscription in the context broker using the default value (1).

taliaga:
Closes #177

### DIFF
--- a/src/utils/subscription_dsl.py
+++ b/src/utils/subscription_dsl.py
@@ -66,7 +66,7 @@ def metadata(value):
 
 
 def throttling(value):
-    return node('throttling', value if value else 1)
+    return node('throttling', value if value is not None else 1)
 
 
 def first_of(*xs):
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -2,6 +2,9 @@
 
 ## 0.6.1
 
+- Time_index enhancement (#172)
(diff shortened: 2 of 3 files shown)


src/utils/subscription_dsl.py
```diff
@@ -66,7 +66,7 @@ def metadata(value):
 
 
 def throttling(value):
-    return node('throttling', value if value else 1)
+    return node('throttling', value if value is not None else 1)
```
c0c0n3: do we need some validation here? e.g. what if the client passes in: `"12", "12.3", 12.3, "nonsense"`. I don't think we've done validation in other places though, so it's probably best to leave it out for now and run a mammoth validation task sometime in the future for all input needing validation?

taliaga: When adding the test I thought the same. Not sure yet if that will bring more value than headaches (for this example of the subscription at least). Probably better to be decided with a mammoth task as you mention, let's discuss it later.

chicco785:
subscription from QL are -10 priority, given that you can (and should)
actually use orion for that.

Issue #186, opened 2019-05-08
Review CI build time


taliaga:
Builds in CI are starting to timeout, plus there are some legacy stuff (packages in docker image) that can be improved. Adjust as necessary.

PR #187, opened 2019-05-08
Fix GH #186

taliaga:


chicco785:
LGTM

taliaga:
Tried to add sql tests but those are broken. They should be checked in https://github.com/smartsdk/ngsi-timeseries-api/issues/188

Issue #185, opened 2019-05-06
Possible fix for multiple data elements in notifications not supported yet


ohylli:
I was attempting to find the most efficient way to send a lot of updates to QuantumLeap and so decided to use the notify api end point directly. Then I noticed that it only supports notifications containing only one element in the data array. If more than one element is given a response with status 400 and content "Multiple data elements in notifications not supported yet" is returned. I started to explore the code and found out that adding support for multiple elements was quite simple. I wonder if there is something I am missing here and thus decided to first create an issue about this before making a full pull request with tests and everrything. 

My modifications were based on the [latest commit](https://github.com/smartsdk/ngsi-timeseries-api/tree/91bbf74e18d8f496ac30878a0905231fd9bb36ec)
in the v0.6.1rc branch. I modified the notify function in [reporter.py](https://github.com/smartsdk/ngsi-timeseries-api/blob/91bbf74e18d8f496ac30878a0905231fd9bb36ec/src/reporter/reporter.py)
 to loop through every element in the data array to perform all operations done to the single payload before (validation, adding the time index etc.). This list of entity updates is then given to the insert method of the CrateTranslatorInstance. In [crate.py](https://github.com/smartsdk/ngsi-timeseries-api/blob/91bbf74e18d8f496ac30878a0905231fd9bb36ec/src/translators/crate.py)
 I modified the `_preprocess_values` method to add None to the values list if the entity update does not have a value for an attribute for which some other entity update of the same type in the notification had a value. This then seems to cause null to be inserted to the database as the value for the attribute. The current implementation throws an NotImplementedError with the message "Seems like not all entities of same type came with the same set of attributes".

With these modifications notifications with multiple updates seem to work all right: the data is saved correctly to crate and returned correctly when queried through the QuantumLeap API. All tests also pass except the one that specifically tests that multiple elements in the notification data gives a response with status 400. Since this was so easy because the implementation almost is there already I started to wonder why it was not yet implemented and if there actually is something wrong with my solution and the correct way to fix this is more complicated.

chicco785:
dear @ohylli 
thanks for looking into quantum leap :)
most probably it was not implemented since at afaik, orion return 1 data element per notification, but i may be wrong :) @taliaga ?

cheers,
federico

chicco785:
@ohylli we discussed in our weekly sprint. we assumed multiple data notification where never issued by orion. which is not the case in a special case, the initial notification:
[...]

PR #191, opened 2019-05-13
Add support for multiple data elements in notifications

ohylli:
The notify API can now process notifications whose data array contains more than one element.
Previously it returned an error message with the HTTP status 400.
reporter.reporter.notify method now loops through each data element
performing the operations previously done to the single element such as adding the time index and validating.
translators.crate.CrateTranslator._preprocess_values method now adds None to the values list if an entity update doesnot have a value for a given attribute.
Before it threw a KeyError exception. This allows _insert_entities_of_type to successfully insert
entities that have different attributes.

Tests were modified to test for this new feature:
- Test data was added to conftest
- test_notify.test_no_multiple_data_elements was renamed and modified to test the notify API with the test data.
- test_insert_same_type_entities_with_different_attrs test was added to test_crate.
   It uses the test data to check that the above described modification to CrateTranslator works.
- Utils.common.check_notifications_record method was modified to use None if there is no value for an attribute in an entity update.
  This was required for the method to work in the test_insert_same_type_entities_with_different_attrs test.

This pull request solves issue #185

### DIFF
--- a/src/conftest.py
+++ b/src/conftest.py
@@ -147,6 +147,53 @@ def entity():
     }
     return entity
 
+@pytest.fixture
+def sameTypeEntitiesWithDifferentAttrs():
+    """
+    Two updates for the same entity with different attributes.
+    The first update has temperature and pressure but the second update has only temperature.
+    """
+    entities = [
+        {
+            'id': 'Room1',
+            'type': 'Room',
+            'temperature': {
+                'value': 24.2,
+                'type': 'Number',
+                'metadata': {
+                    'dateModified': {
+                        'type': 'DateTime',
+                        'value': '2019-05-09T15:28:30.000'
+                    }
+                }
+            },
+            'pressure': {
+                'value': 720,
+                'type': 'Number',
+                'metadata': {
+                    'dateModified': {
+                        'type': 'DateTime',
+                        'value': '2019-05-09T15:28:30.000'
+                    }
+                }
+            }
+        },
+        {
+            'id': 'Room1',
+            'type': 'Room',
+            'temperature': {
+                'value': 25.2,
+                'type': 'Number',
+                'metadata': {
+                    'dateModified': {
+                        'type': 'DateTime',
+                        'value': '2019-05-09T15:29:30.000'
+                    }
+                }
+            }
+        }
+    ]
+    return entities
 
 @pytest.fixture
 def air_quality_observed():
--- a/src/conftest.py
+++ b/src/conftest.py
@@ -148,7 +148,7 @@ def entity():
     return entity
 
 @pytest.fixture
-def sameTypeEntitiesWithDifferentAttrs():
+def sameEntityWithDifferentAttrs():
     """
     Two updates for the same entity with different attributes.
     The first update has temperature and pressure but the second update has only temperature.
--- a/src/reporter/reporter.py
+++ b/src/reporter/reporter.py
@@ -94,7 +94,7 @@ def _validate_payload(payload):
     # (i.e, the changed value)
     attrs = list(iter_entity_attrs(payload))
     if len(attrs) == 0:
-        log().warning("Received notification without attributes " +
+        log().warning("Received notification containing an entity update without attributes " +
                       "other than 'type' and 'id'")
 
     # Attributes should have a value and the modification time
@@ -102,7 +102,7 @@ def _validate_payload(payload):
         if not has_value(payload, attr):
             payload[attr].update({'value': None})
             log().warning(
-                'Payload is missing value for attribute {}'.format(attr))
+                'An entity update is missing value for attribute {}'.format(attr))
 
 
 def notify():
@@ -115,26 +115,25 @@ def notify():
                'content.', 400
 
     payload = request.json['data']
-    if len(payload) > 1:
-        return 'Multiple data elements in notifications not supported yet', 400
-    payload = payload[0]
 
     log().info('Received payload: {}'.format(payload))
 
-    # Validate notification
-    error = _validate_payload(payload)
-    if error:
-        return error, 400
-
-    # Add TIME_INDEX attribute
-    payload[CrateTranslator.TIME_INDEX_NAME] = \
-        select_time_index_value_as_iso(request.headers, payload)
-
-    # Add GEO-DATE if enabled
-    add_geodata(payload)
-
-    # Always normalize location if there's one
-    normalize_location(payload)
+    # preprocess and validate each entity update
+    for entity in payload:
+        # Validate entity update
+        error = _validate_payload(entity)
+        if error:
+            return error, 400
+    
+        # Add TIME_INDEX attribute
+        entity[CrateTranslator.TIME_INDEX_NAME] = \
+            select_time_index_value_as_iso(request.headers, entity)
+    
+        # Add GEO-DATE if enabled
+        add_geodata(entity)
+    
+        # Always normalize location if there's one
+        normalize_location(entity)
 
     # Define FIWARE tenant
     fiware_s = request.headers.get('fiware-service', None)
@@ -149,7 +148,7 @@ def notify():
 
     # Send valid entities to translator
     with CrateTranslatorInstance() as trans:
-        trans.insert([payload], fiware_s, fiware_sp)
+        trans.insert(payload, fiware_s, fiware_sp)
 
     msg = 'Notification successfully processed'
     log().info(msg)
--- a/src/reporter/tests/test_notify.py
+++ b/src/reporter/tests/test_notify.py
@@ -215,27 +215,16 @@ def test_geocoding(notification, clean_mongo, crate_translator):
     assert float(lat) == pytest.approx(24.9412167, abs=1e-2)
 
 
-def test_no_multiple_data_elements(notification, clean_mongo, clean_crate):
-    second_element = {
-                        'id': 'Room2',
-                        'type': 'Room',
-                        'temperature': {
-                            'type': 'Number',
-                            'value': 30,
-                            'metadata': {
-                                'dateModified': {
-                                    'type': 'DateTime',
-                                    'value': '2017-06-20T11:46:45.00Z'
-                                }
-                            }
-                        }
-                    }
-    notification['data'].append(second_element)
+def test_multiple_data_elements(notification, sameTypeEntitiesWithDifferentAttrs, clean_mongo, clean_crate):
+    """
+    Test that the notify API can process notifications containing multiple elements in the data array.
+    """
+    notification['data'] = sameTypeEntitiesWithDifferentAttrs
+    print(json.dumps(notification))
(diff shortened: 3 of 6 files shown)


src/conftest.py
```diff
@@ -147,6 +147,53 @@ def entity():
     }
     return entity
 
+@pytest.fixture
+def sameTypeEntitiesWithDifferentAttrs():
+    """
```
chicco785: i wonder why only testing same entity id?

src/translators/crate.py
```diff
@@ -313,10 +313,9 @@ def _preprocess_values(self, e, col_names, fiware_servicepath):
                         values.append([float(lon), float(lat)])
                     else:
                         values.append(e[cn]['value'])
-                except KeyError as e:
-                    msg = ("Seems like not all entities of same type came "
-                           "with the same set of attributes. {}").format(e)
-                    raise NotImplementedError(msg)
+                except KeyError:
+                    # this entity update does not have a value for the column so use None which will be inserted as NULL to the db.
+                    values.append( None )
```
chicco785: not sure about the implications of this change? @taliaga @c0c0n3 ?

src/utils/common.py
```diff
@@ -234,7 +234,9 @@ def check_notifications_record(notifications, records):
             continue
 
         r_values = record[a]['values']
-        n_values = [e[a]['value'] for e in notifications]
+        # collect values for the attribute a from the entities in the notification
+        # if an entity does not have a value for the attribute use None
+        n_values = [e.get( a, { 'value': None } )['value'] for e in notifications]
```
chicco785: so basically, this is the place where "none" is entered if an attribute value is missing.

i think a specific test to validate this change is needed. also this change may be better to be in a separate pr
@taliaga @c0c0n3 ?

src/conftest.py
ohylli: No special reason for testing with same entity id. It just happens to match my use case for this feature. Different ids would work as well although test_insert_same_type_entities_with_different_attrs would require some changes since query would return 2 records not one and utils.common.check_notifications_record requires that all entities have the same id and there is only one record.

src/translators/crate.py
c0c0n3: I think this is actually a good change. In fact, you get a `KeyError` when accessing a key that's not in a dictionary. In our case that can happen only if the attribute dictionary has no `value` key, so setting it to `None` which will ultimately result in a DB `null` for the corresponding field is actually consistent with what we've done in PR #122. Also, PR #122 should make this code block unreachable actually, in other words I think the `except` clause is dead code!

src/utils/common.py
c0c0n3: As far as I can tell, the setting of missing attribute values to `None` happens in the `reporter`

https://github.com/smartsdk/ngsi-timeseries-api/pull/122/files/a78eb256fbe307599ce1c6fef2e0f9623f9f3c50#diff-f796da60daee45177c252e52fc78b4aaR96:

Again, this is a change we brought in with PR #122---see my comment above.
On the other hand this block of code @ohylli changed is part of the `check_notifications_record` function which is only called from test code, so if all tests pass, this must be a good change, right? However, in principle this line shouldn't be needed, again because of PR #122. So I suggest rolling back the changes to this file, unless doing so causes some tests to fail, in which case we'd have to figure out why.

src/translators/crate.py
ohylli: With my modifications this except clause is not dead code. Its required for example when inserting the new sameTypeEntitiesWithDifferentAttrs test entities I added to src/conftest. There the first entity has attributes temperature and pressure and the second only temperature. This is different from the situation in pull request #122 where the attribute is present but has no value where as here the attribute pressure is completely missing from the entity update. So when _preprocess_values processes the latter entity update the KeyError exception is raised and None is added as the value for pressure. Of course None could be added earlier in src/reporter/reporter.py as its done with missing values but I did it here since it was a simple change.

src/utils/common.py
ohylli: This modification is required by the new Test_insert_same_type_entities_with_different_attrs test I added where I use this method. This test uses the new test data I added to src/conftest. As I said in my previous comment about my modification to _preprocess_values the situation with this modification is also different from pr #122 since here the attribute is completely missing instead of only the value missing.

src/translators/tests/test_crate.py
```diff
@@ -29,6 +29,21 @@ def test_insert_entity(translator, entity):
 
     check_notifications_record([entity], loaded_entities)
 
+def test_insert_same_type_entities_with_different_attrs( translator, sameTypeEntitiesWithDifferentAttrs ):
```
taliaga: Ok, good, this is one of the tests I mentioned above. There's a catch though at assertion time, see my comment on ```common.py```

src/utils/common.py
taliaga: I understand this is a needed change because now with this PR the symmetry of the entities (in terms of set of attributes) is no longer hold, because now an `a` (attribute) of entity `e1` (entity) might not exist in another `e2`.

The catch I mentioned earlier is, under this assumption of symmetry, the `check_notifications_record` method has been taking the set of attributes from the first record... https://github.com/smartsdk/ngsi-timeseries-api/blob/f639c6c3a5c4d774b59db8e16716b0616ca747ab/src/utils/common.py#L221. To make sure you test all attributes (their values and Nones), you should instead take the complete set of attributes (considering all records).

Thanks in advance for your patience with our comments. Since this PR is changing a critical assumption, we want to make sure we don't oversee places where this symmetry break could cause more problems (specially those that do not break tests)

On a side note,
- Maybe slightly cleaner: ```n_values = [e[a]['value'] if a in e else None for e in notifications]```
- No need for a separate PR IMO

src/reporter/reporter.py
```diff
@@ -115,26 +115,25 @@ def notify():
...
-    add_geodata(payload)
-
-    # Always normalize location if there's one
-    normalize_location(payload)
+    # preprocess and validate each entity update
+    for entity in payload:
+        # Validate entity update
+        error = _validate_payload(entity)
+        if error:
+            return error, 400
```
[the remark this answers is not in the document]
ohylli: This seems not to be possible or at least not simple. Only things _validate_payload reports errors about are missing type or id but Connexion alrady checks them before an gives its own error which only reports the issue it noticed first. I know nothing about Connexion so I don't know if this could be some how done with it.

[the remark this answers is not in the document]
c0c0n3: also, we'll have to think about what to do if validation for a record fails. What if we have a batch of 10 and two entities fail validation? Are we going to ditch the whole lot?! Or should we rather accept the 8 that are valid and return an error message indicating partial failure, possibly with an indication of which entity failed validation and why?

src/reporter/tests/test_notify.py
```diff
@@ -215,27 +215,16 @@ def test_geocoding(notification, clean_mongo, crate_translator):
...
-                            'metadata': {
-                                'dateModified': {
-                                    'type': 'DateTime',
-                                    'value': '2017-06-20T11:46:45.00Z'
-                                }
-                            }
-                        }
-                    }
-    notification['data'].append(second_element)
+def test_multiple_data_elements(notification, sameTypeEntitiesWithDifferentAttrs, clean_mongo, clean_crate):
```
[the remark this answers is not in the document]
c0c0n3: yes, definitely.

src/translators/crate.py
c0c0n3: oh right! thanks for clarifying I missed that.

src/reporter/tests/test_notify.py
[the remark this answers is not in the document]
ohylli: Do you mean tests for the notify API i.e. reporter.notify or data base insert i.e. CrateTranslator.insert. For the latter there I already added one test as you noticed. For testing elements with different entity types does the existing test_query_all in test_crate work or is a new test with entities of different types including same type entities with different attributes required.

src/utils/common.py
ohylli: Is this required for this pull request since all tests including my new one gives one record for this. check_notifications_record function also has an assert for checking that there is only one record.

ohylli: Of course if we end up creating more tests for this pr then  check_notifications_record  probably will require this modification.

src/conftest.py
```diff
@@ -147,6 +147,53 @@ def entity():
     }
     return entity
 
+@pytest.fixture
+def sameTypeEntitiesWithDifferentAttrs():
```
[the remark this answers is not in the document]
ohylli: I changed the name and the related test name since it is more clear.

src/reporter/tests/test_notify.py
```diff
@@ -215,27 +215,16 @@ def test_geocoding(notification, clean_mongo, crate_translator):
...
+    """
+    Test that the notify API can process notifications containing multiple elements in the data array.
+    """
+    notification['data'] = sameTypeEntitiesWithDifferentAttrs
+    print(json.dumps(notification))
```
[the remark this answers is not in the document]
ohylli: I removed it.

ohylli:
@chicco785, @c0c0n3, @taliaga I think I have addressed all of your comments with commits or my own comments. Since you have not responded for a few days I wanted to check if you are expecting something from me or if you just have not yet had time to look into this which I naturally understand.

c0c0n3:
@ohylli sorry for the slow reply :-)
we actually talked about this PR yesterday and are thinking of merging it in. You can ignore my comments about error handling and batch processing, since we decided to implement those features in a separate PR. For the rest, we're all very happy with your work and would like to thank you very much for taking the time to do this.

SourabhChourasiya-NEC:
Hi @ohylli and @c0c0n3 Sir,

I am also facing the same issue in our project and so I am interested in this feature to be merged as soon as possible.
Could please merge this issue or let me know in case any contribution is required. I am happy to help. 

Thanks

c0c0n3:
Hi @SourabhChourasiya-NEC, thanks for offering to help :-)
We should be able to work on this next week and will definitely let you know if we need an extra pair of hands, thanks!

src/utils/common.py
taliaga: no problem, we can take care of that later.

src/reporter/reporter.py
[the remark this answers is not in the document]
taliaga: Will be covered by https://github.com/smartsdk/ngsi-timeseries-api/issues/193

src/reporter/tests/test_notify.py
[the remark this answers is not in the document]
taliaga: Both. In your use-case, are you inserting multiple entities always of the same type or sometimes mixed?

BTW, we'll not block this PR for this missing test, but it's something that needs to be added.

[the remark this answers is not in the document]
ohylli: In my use-case the entities are always of the same type.

PR #198, opened 2019-05-24
Support for NGSI_GEOPOINT for null values

pooja1pathak:
While executing API {ql_ip}:{ql_port}/v2/entities/{entity_id} to get all attributes of the entity having **"null"** values for the attribute of type **"geo:point"** an error is shown as follows: 
**"Something went wrong with QL. Error: 'NoneType' object is not iterable"**.

After investigation, I have found that this error occurs because of the code which insert value of longitude and latitude with the None value.

To remove this error, I have added a check for the None value when type is NGSI_GEOPOINT.


chicco785:
@pooja1pathak to ensure that similar issue won't happen in the future, would you main creating a test for this issue?

pooja1pathak:
Hi @chicco785 , 
As suggested by you, I have created test case for the same.

taliaga:
thanks!

Issue #213, opened 2019-06-18
Flatten JSON in query results


c0c0n3:
**Current query result format**
We've been using the following format for query results: `{ data: { ...query result...} }`. The `data` attribute adds a level of nesting which isn't needed.

**Proposed solution**
Remove the `data` attribute from the JSON returned by the queries and put the fields that make up the query result at the top level as in e.g. `{ entityId: ..., index: ..., values: ...  }`

**Backward compatibility**
The proposed change **breaks API backward compatibility** so it will have to be managed accordingly at API and release level---e.g. new API version, advertise change, etc.

c0c0n3:
closed by #214.

PR #214, opened 2019-06-18
Flatten JSON in query results

c0c0n3:
This PR implements the flattening JSON query responses as outlined in #213 and additionally changes the array type of values returned by the queries from an array of numbers to an array of any type. The reason for the latter change is that attribute values may not only be numbers, but any of the valid types permitted by the NGSI spec.

### Important: Backward compatibility
The implemented changes break API backward compatibility so it will have to be managed accordingly at API and release level---e.g. new API version, advertise change, etc.


c0c0n3:
have also updated the spec in the sdk repo to reflect the change to the value array type, see:

https://github.com/orchestracities/sdk/commit/49885dbc8d8cac23e55360db1260823640ad9592

PR #218, opened 2019-06-21
Adjust spec version and add missing element

taliaga:


specification/quantumleap.yml
```diff
@@ -83,6 +83,7 @@ definitions:
 
   ValuesArray:
     type: array
+    items:
```
c0c0n3: This is hairy. The Swagger 2 spec basically says it should be the same as `items` in the JSON schema:

* https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#schemaObject

but the JSON schema they refer to is v4:

* https://tools.ietf.org/html/draft-zyp-json-schema-04

which seems to say nothing about the meaning of `items` without a type. The Swagger 3 spec says: 

> Unlike JSON Schema, the items keyword is required in arrays. The value of items is a schema that describes the type and format of array items

which seems to imply you have to specify a type, but then again it isn't saying it explicitly.

What should we do then? Does this work in SwaggerHub? And does it work in Quantum Leap with response validation turned on?

chicco785: hi @c0c0n3, i think for openapi solution is very easy:
https://swagger.io/docs/specification/data-models/data-types/

check "Mixed-Type Arrays":

```
# ["foo", 5, -2, "bar"]
type: array
items:
  oneOf:
    - type: string
    - type: integer
```
i suspect we could have also array, object or whatever there.

another alternative if this is not enough, is refer to a schema that is "Any Type" (check the section in the same link above):

```
components:
  schemas:
    AnyValue: {}
```

c0c0n3: I'd use `AnyValue` just to get the ball rolling when we make the switch to Swagger 3/Open API. Then I'd also like to use the `oneOf` attribute, but that will have to be kept in synch with the types we accept in the notifications which might require a bit more of thinking since we'll have to review the mapping we do on inserting entities.

In the meantime, since we have Swagger 2 but will soon update to 3, for my money this PR can go in as is as long as it:

* works in SwaggerHub
* works in QL w/ response validation **turned off**

@taliaga does it? :-)

c0c0n3: Cool, so it works on SwaggerHub and in QL with response validation off. Happy days!

taliaga: I agree with your comments, we need to review the actually supported array types with QL (and either prevent or support mixed types).

The purpose of this PR was to align the versioning as requested. In the same go, I wanted to leave it clean of warnings (swagger validator was complaining about an array without items). Empty items was OK but I've checked now again and according to https://swagger.io/docs/specification/data-models/data-types/ it should have a {}. 

This is just to keep consistency and remove warnings. As mentioned, we'll need to review the types altogether.

PR #221, opened 2019-07-02
Coverage tool integration with travis

SourabhChourasiya-NEC:
I have integrated Coveralls with Travis for fiware/quantumleap. 

**There are below mentioned guidelines which should be followed by the repository owner/member:**

**For Coveralls:**
```
1. Go to https://coveralls.io/sign-in
2. Click on "GITHUB SIGN IN"
3. On Left end side go to menu bar.
4. Click on "ADD REPOS"
5. Find the repository name and enable its integration with coveralls by on-off toggle switch.
6. Click on the link "DETAILS" associated with the repo and go to settings available at left upper corner.
7. Under section "README BADGE" click on dropdown "EMBED" button and copy the content of "MARKDOWN" for further process.
8. Copy the repo token from variable "REPO TOKEN" for further process.
9. Set the "COVERAGE THRESHOLD FOR FAILURE".
10.Set the "COVERAGE DECREASE THRESHOLD FOR FAILURE".
11.Click on "SAVE CHANGES"
```

**On Travis:**
```
As "smartsdk/ngsi-timeseries-api" repository already sink with travis CI and we should only update the configuration at travis side.
1. Go to repositories section available at https://travis-ci.org/account/repositories.
2. Click on settings for "smartsdk/ngsi-timeseries-api" repository.
3. Under "Environment Variables" section, please add a new variable with below details: 
   Name : COVERALLS_REPO_TOKEN
   Value: {repo_token}  (paste the REPO TOKEN here which is copied earlier).
```

**On Github:**
```
1. Go to https://github.com/smartsdk/ngsi-timeseries-api/blob/master/README.md
2. Edit the README.md 
3. Paste the "MARKDOWN" content here for adding a badge of coveralls.
```


setup_dev_env.sh
```diff
@@ -23,4 +27,4 @@ export RETHINK_HOST=$LH
 
 export REDIS_HOST=$LH
 
-pipenv shell
+[[ "$LH" != "192.0.0.1" ]] || pipenv shell
```
taliaga: So, why not using the pipenv shell also in Travis?

src/geocoding/tests/run_tests.sh
```diff
@@ -5,8 +5,10 @@ docker build -t quantumleap ../../../
 docker-compose up -d
 sleep 8
 
-docker run -ti --rm --network tests_geocodingtests quantumleap pytest geocoding/
+cd ../../../
+pytest src/geocoding/ --cov-report= --cov-config=.coveragerc --cov-append --cov=src/
```
taliaga: Any reason for running pytest outside the container? Is this way easier to integrate with coverage tool maybe?

Note for this pytest command to work in the local environment, we need at least to install the plugins used by this command, and make sure we're using the same pytest version. Could you put those in the pipenv file maybe (instead of installing them in the travis script)? This way we make sure we always install the same set of tools, everywhere. For the time being, we don't mind having the same pipenv file for all environments (testing, dev or prod).

Maybe there's something else that needs to be done locally for the coverage to work (or to be disabled if so wanted)?

And maybe at this point the run_tests scripts are no longer worth much, but that we can discuss other day in another PR :)

src/tests/run_tests.sh
```diff
@@ -11,10 +11,13 @@ tot=0
 QL_IMAGE=${QL_PREV_IMAGE} docker-compose -f ../../docker/docker-compose-dev.yml up -d
 sleep 10
 
+ORION_HOST=`docker inspect -f '{{range .NetworkSettings.Networks}}{{.IPAddress}}{{end}}' docker_orion_1`
```
taliaga: I wonder if you didn't need to do this also for the tests in the reporter package. Note that in QL not all the tests are using the `docker/docker-compose-dev/yml` file (where you are giving a name to these containers).

src/tests/run_tests.sh
```diff
@@ -23,21 +26,17 @@ QL_IMAGE=quantumleap docker-compose -f ../../docker/docker-compose-dev.yml up -d
 sleep 10
 
 # Backwards Compatibility Test
-docker run -ti --rm --network docker_default \
-    -e ORION_URL="http://orion:1026" \
-    -e QL_URL="http://quantumleap:8668" \
-    quantumleap pytest tests/test_bc.py
+cd ../../
```
taliaga: why is this change of directories needed?  If it's just for the location of the cov-config maybe it's easier to adjust its reference (--cov-config=../../)

src/reporter/tests/run_tests.sh
```diff
@@ -5,8 +5,10 @@ docker build -t quantumleap ../../../
 docker-compose up -d
 sleep 12
 
-docker run -ti --rm --network tests_reportertests quantumleap pytest reporter/tests
+cd ../../../
+pytest src/reporter/ --cov-report= --cov-config=.coveragerc --cov-append --cov=src/
```
taliaga: also, if coverage adds too much overhead for the local day-to-day development, would you be interested in having a simple way to disable coverage for the runs?

setup_dev_env.sh
SourabhChourasiya-NEC: According to my understanding "pipenv shell" command is used to install quantumleap as a source for running of app.py file. In case of travis, all the services are running as a containerized enviroment, so there is no need to run app.py file for running quantumleap service in case of travis.

src/geocoding/tests/run_tests.sh
SourabhChourasiya-NEC: When the test cases are running by pytest, we have generate the coverage report by pytest-cov. 
Quantumleap code and test cases are in same directory for each submodules (translator, reporter, geocoding etc.).
In previous method of test case execution, following are the command which execute from run_tests.sh script for each submodule:
docker-compose up -d
docker run -ti --rm --network tests_translatorstests quantumleap pytest translators/tests
docker-compose down -v

Through above commands, it will create report for one submodule and if execution of next test case for another submodule, will proceed then the previous report will not be available at that time as docker down command will execute with end of each test case and due to this final report will not be cover the complete result.
Now if we keep the report creation outside of container then we will be able to get the expected report. And we have followed the followed the same docker-compose.yml files for each submodules but test cases are running outside the container.

Yes, i will update the Pipfile for all the dependencies.

src/tests/run_tests.sh
SourabhChourasiya-NEC: The "container_name" variable was used to fetch the IP address for inter container communication. Now we are updating the script file for getting the IP address dynamically by based on the port on which Orion and Quantumleap are running.

src/tests/run_tests.sh
SourabhChourasiya-NEC: The work directory location is used for storing the common coverage report ".coverage". It is required to run pytest from work directory and it will append the result on existing coverage report file for each and every submodules.

.travis.yml
```diff
@@ -9,10 +9,19 @@ services:
 
 install:
   - docker --version
+  - pip install coveralls
+  - pip install pytest-cov
```
c0c0n3: as you guys pointed out, the two lines above:

    - pip install coveralls
    - pip install pytest-cov

should be removed from here  and  Python deps should be added to the Pipfile. In the `script` section below you `source setup_dev_env.sh` which in turn starts a `pipenv shell`, so all deps will be available in the Travis environment as well as when we run it locally. Thanks!!!

setup_dev_env.sh
c0c0n3: see my comment above about `pipenv shell`. I think we should run it in Travis too so we should actually rollback this line with the conditional statement?

src/reporter/tests/run_tests.sh
c0c0n3: Yes, this would definitely be a nice to have!

src/tests/run_tests.sh
c0c0n3: yep, I agree we should check all the other test cases too, though I'd have expected them to fail...

taliaga: sure, I understood why the naming was needed, I was just curious to see it wasn't needed in at least the docker-compose used in the reporter tests. Anyway, nevermind.

src/geocoding/tests/run_tests.sh
taliaga: OK! can you make sure the tests output is still visible after a pytest run both locally and on travis by default?

c0c0n3:
@SourabhChourasiya-NEC this is great thank you sooo much, from all of us, for all the hard work you put in. There's just one little snag though. Like @taliaga mentioned earlier, before we can merge this PR, we need to find a way to display `pytest` output to `stdout/stderr`. We have to have this when we run the tests locally to see immediately what went wrong and also on Travis, otherwise it could be overly difficult to find out why a test run failed on Travis.

SourabhChourasiya-NEC:
Hi @c0c0n3, @taliaga Sir,

I have checked on local as well as Travis build time, it is showing the pytest output on stdout/stderr. If any of test case is failing it will show on the console, so a user can debug it.

Also if the user wants to see the coverage result on his local setup, user can use command ```coverage report ``` to check the coverage report locally.

Please verify at your end if any other thing is remaining you can let me know so we can close this PR as soon as possible.

Thanks

setup_dev_env.sh
SourabhChourasiya-NEC: About ```pipenv shell```, I have investigated and found an issue on [GitHub](https://github.com/pypa/pipenv/issues/3500). Please look into this and provide your opinion on it.

c0c0n3: @SourabhChourasiya-NEC thanks for pointing it out!
There's something I should clear out though. My comment earlier was misleading: we actually never run this script on Travis, we only use it to set up a *local* test environment on our boxes. But I suppose we can leave this file as is, it looks like your changes actually make the script more robust, thanks!

c0c0n3:
Hi @SourabhChourasiya-NEC,

Sorry for this PR dragging on, we're swamped with work at the moment and @taliaga is away on paternity leave :-)

I think the PR is fine the way it is, but there's still that issue of not being able to see the test output locally. I checked out your branch, ran `sh setup_dev_env.sh` followed by `sh run_tests.sh` and couldn't see any test output. On closer inspection, I noticed this message buried under tons of other text:

    pytest: command not found

So it looks like the reason we couldn't see any test output is that tests don't actually get run! I leave this in your capable hands, but I think now that we run tests outside of the QL container, pytest needs to be added to the PATH?

SourabhChourasiya-NEC:
Hi @c0c0n3 Sir,

As mentioned in the ```.travis.yml``` file before running  ```source setup_dev_env.sh```  and ```sh run_tests.sh```, it is important to run ```pipenv install``` for installing all the dependency locally.

Could you please run below-mentioned command and try it again:
1. pipenv install
2. source setup_dev_env.sh
3. sh run_tests.sh

If you facing any other issue, please let me know.

Thanks

c0c0n3:
@SourabhChourasiya-NEC yep, I had to run `pipenv install`, now it works like a charm! Sorry for being a dumb @**! The build's failing after you merged `master` into your branch though. The reason is that I added some new tests in #231 that probably need some massaging to run in the new test env. But don't worry about that, I'll see if there's a way to force the merge---GitHub blocked the merge. Then I'll adapt my tests after the merge. Again, thank you sooooo much for all the hard work you put in, much appreciated!!!

SourabhChourasiya-NEC:
Hi @c0c0n3 SIr,

I am glad to know that it is working fine now.

Thanks

PR #226, opened 2019-07-18
Disable osm check by default

taliaga:


taliaga:
Change to avoid k8s probes abuse osm calls.

Issue to fix failing tests: https://github.com/smartsdk/ngsi-timeseries-api/issues/227

c0c0n3:
@taliaga there's one thing beyond me though. How is the change in the health function going to help with the failing tests? I was expecting those tests to pass in this branch, but they are not?

taliaga:
@c0c0n3 connection to osm in geocoder tests was timing out.

PR #243, opened 2019-08-08
Timescale backend teething troubles, take 2

c0c0n3:
This PR contains some fixes and enhancements to the recently released Timescale backend.

* **Fix**. SSL support for QuantumLeap DB setup container (da38a3c). When running `quantumleap-db-setup.py` inside the setup container, we'll automagically use SSL if available on the server or fall back to a plain TCP connection otherwise.
* **New**. Logging of Postgres connection data (975af69). When using the Timescale backend, we log any connection parameter read from the environment, making sure passwords never get written to the logs.
* **Fix**. Sync up columns when exporting from Crate (46a9f3d). It could happen that the set of columns in Crate at the time of exporting data (using `crate-exporter.py`) is larger than that of Postgres. For instance, say you have a device that measures temperature and wind speed. Sometimes the device sends both measurements while some other times just one of the two. If you have collected data in Crate for a long period of time, you'll likely have both columns but when you switch over to Timescale, you may initially only have one column since the device has only sent e.g. temperature data so far.


src/translators/tests/test_pg_conn_data.py
```diff
@@ -0,0 +1,58 @@
...
+    assert_conn_param(getter, env_var_name, '')
+
+
+@pytest.mark.parametrize('getter, env_var_name, env_value, expected_param', [
+    (get_host, 'POSTGRES_HOST', ' my.host ', 'my.host'),
+    (get_port, 'POSTGRES_PORT', ' 5432 ', 5432),
+    (get_ssl, 'POSTGRES_USE_SSL', ' no ', False),
+    (get_ssl, 'POSTGRES_USE_SSL', ' yes ', True),
+    (get_db_name, 'POSTGRES_DB_NAME', 'quantumleap\n', 'quantumleap'),
+    (get_db_user, 'POSTGRES_DB_USER', '\tquantumleap', 'quantumleap'),
```
chicco785: i am just curious, why `\t`?

src/translators/tests/test_pg_conn_data.py
```diff
@@ -0,0 +1,58 @@
...
+def test_param_default(getter, env_var_name):
+    assert_conn_param(getter, env_var_name, '')
+
+
+@pytest.mark.parametrize('getter, env_var_name, env_value, expected_param', [
+    (get_host, 'POSTGRES_HOST', ' my.host ', 'my.host'),
+    (get_port, 'POSTGRES_PORT', ' 5432 ', 5432),
+    (get_ssl, 'POSTGRES_USE_SSL', ' no ', False),
+    (get_ssl, 'POSTGRES_USE_SSL', ' yes ', True),
+    (get_db_name, 'POSTGRES_DB_NAME', 'quantumleap\n', 'quantumleap'),
```
chicco785: i am just curious, why `\n`?

src/translators/tests/test_pg_conn_data.py
c0c0n3: I've seen automatically generated settings of env vars where some extra white space seeped in---I don't think this can possibly happen with our charts though.

src/translators/tests/test_pg_conn_data.py
c0c0n3: I've seen automatically generated settings of env vars where some extra white space seeped in---I don't think this can possibly happen with our charts though.

src/utils/tests/test_cfgreader.py
```diff
@@ -0,0 +1,153 @@
...
+    assert var.read(value) == def_val
+
+
+@pytest.mark.parametrize('value, expected', [
+    (' 5', False), ('\t5 ', False), ('5\n', False), (' 5 ', False),
+    ('false', False),
+    ('t', True), (' t', True), ('\tt ', True), ('t\n', True), (' t ', True),
+    ('T', True), (' T', True), ('\tT ', True), ('T\n', True), (' T ', True),
+    ('1', True), (' 1', True), ('\t1 ', True), ('1\n', True), (' 1 ', True),
+    ('yes', True), (' yes', True), ('\tyes ', True), ('yes\n', True),
```
taliaga: out of curiosity, who defines these valid options for `true`? E.g., why is `y` not there or why there shouldn't be?

timescale-container/test/pg-ssl/how-to-generate-certs.md
```diff
@@ -0,0 +1,50 @@
...
+
+**Warning**: The below procedure is only suitable for testing locally,
+don't even think of doing the same for prod!
+
+
+### Create self-signed certificate
+
+Create a simple self-signed certificate for localhost valid for 10 years:
+
+    $ openssl req -new -x509 -days 365 -nodes -text -out server.crt \
```
taliaga: 365 or 3650?

src/utils/cfgreader.py
```diff
@@ -0,0 +1,167 @@
...
+        :return: the parsed value if the variable is set, the variable's
+            default otherwise.
+        """
+        env_value = self.var_store.get(var.name)
+        msg = self.get_log_msg(var, env_value)
+        self.log(msg)
+        return var.read(env_value)
+
+
+class YamlReader:
```
taliaga: I'd use a different name, what kind of yaml reader is this one?. 

Also, on a side note, I'd organise a bit better these ql-config-related files into a package with its specific purpose (config handling) to avoid creating a "utils" blackhole here.

src/utils/cfgreader.py
```diff
@@ -0,0 +1,167 @@
...
+MaybeString = Union[str, None]
+
+
+class EVar:
+    """
+    Defines an interface for environment variable parsers.
+    """
+
+    @staticmethod
+    def has_value(str_rep: MaybeString) -> bool:
```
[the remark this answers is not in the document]
c0c0n3: yep, we have an `has_value` in the reporter, but it's for NGSI attrs.

src/utils/tests/test_cfgreader.py
c0c0n3: I've asked the interwebs for the most common ways to map a truthy string to Python's `True` and the answer was: 'true', 'yes', '1', 't'. This is kinda arbitrary anyway, so should I add `y` too?

timescale-container/test/pg-ssl/how-to-generate-certs.md
c0c0n3: 3650. fixed thanks!!!

src/utils/cfgreader.py
c0c0n3: Yes, `YamlReader` isn't a great choice actually. I'll see if I can come up w/ something better. These classes are generic though, i.e. not tied in anyway to QL which is why I put them into util. Actually I've been (ab-)using the util namespace I reckon, I stashed away in there general purpose code that isn't tied to QL: `cfgreader`, `jsondict`, `kvt`, `streams`, and `timestr`. I'll find a better home for these guys when we refactor the timescale and crate backends.

src/utils/tests/test_cfgreader.py
c0c0n3: added `y` too :-)

PR #248, opened 2019-08-22
V0.7.4rc

taliaga:


src/reporter/reporter.py
```diff
@@ -128,7 +128,7 @@ def notify():
     
         # Add TIME_INDEX attribute
         entity[CrateTranslator.TIME_INDEX_NAME] = \
-            select_time_index_value_as_iso(request.headers, entity)
+            select_time_index_value_as_iso(dict(request.headers), entity)
```
taliaga: This casting here at the entrypoint of our codebase because `request.headers` is one of [these](https://tedboy.github.io/flask/_modules/werkzeug/datastructures.html#EnvironHeaders), which won't go through [this if](https://github.com/smartsdk/ngsi-timeseries-api/blob/f7325320da7835bf60bd76e6e5764131c3cb1dba/src/utils/jsondict.py#L51).

Maybe it's worth changing some strict `isinstance` checks by a `try + except` pairs in [jsondict.py](https://raw.githubusercontent.com/smartsdk/ngsi-timeseries-api/master/src/utils/jsondict.py) to be more open to custom dict-like objects?

chicco785: is there any reason why not using flask  native methods to manipulate the headers? 
I am worried keeping the assumption headers is dict may bring pain in the future.

taliaga: I'm not sure I understand the question as such, but I agree with your concern. 

In fact, on a second thought, I think this "headers" object (regardless of its type) should not even go beyond this `reporter` module deeper in the stack. I'll extract the custom_index name here (if available) and adjust accordingly the methods (and tests) down that path of `select_time_index_value_as_iso`.

Issue #249, opened 2019-09-04
Fix lastN implementation


taliaga:
**Describe the bug**
A query to the lastN values of a specific attribute (e.g. https://{{QL}}/v2/entities/urn:ngsi-ld:NoiseLevelObserved:FOO/attrs/LAeq_m?lastN=5) is not returning the lastN values found in the database.

**To Reproduce**

- Note there is an [integration test](https://github.com/smartsdk/ngsi-timeseries-api/blob/b33a1a5e4321f198ffbc65abe5a43d82ac2bd887/src/reporter/tests/test_1T1E1A.py#L174) covering the lastN feature, but the dataset used there is small (30 items).
This is happening in a deployment with many records. QL is taking the lastN values of the chunked resultset (due to the fact that default LIMIT was reached).

**Expected behavior**
It should return the last 5 records from the time of querying.

**Environment:**
 - Version 0.7.4

**Additional context**
- NOTE Queries use by default ASC ordering by time_index so that responses arrays are a time series "from from past to present". The "lastN" feature was quickly implemented for small datasets, by just taking the lastN values in memory after a query without such lastN was done (trusting reasonable defaults). The fix of this issue is the time to properly implement this feature with an efficient query for this case (like doing a DESC ordering + LIMIT=lastN).

PR #250, opened 2019-09-09
Fix GH-249

taliaga:


src/reporter/tests/test_1T1E1A.py
```diff
@@ -200,6 +200,38 @@ def test_1T1E1A_lastN(reporter_dataset):
     assert_1T1E1A_response(obtained, expected)
 
 
+def test_1T1E1A_lastN_with_limit(reporter_dataset):
```
chicco785: is suppose with offset is not possible, correct? dumb question do we give feedback for "uncorrect" queries ?

chicco785: is this affecting only 1T1E1A?

taliaga: > is suppose with offset is not possible, correct? dumb question do we give feedback for "uncorrect" queries?

Which incorrect queries are you thinking specifically? It'd be weird to specify both offset and lastN (like lastN is so big you need to paginate the response), but I believe still valid, no?

> is this affecting only 1T1E1A?

Good question. Also for 1T1ENA, for which I can add a test now. For the more general endpoints (1TNE1A, 1TNENA, etc.) I believe lastN was not fully defined / implemented yet. We could discuss this tomorrow.

chicco785: As incorrect I meant, combinations of parameters that should not be allowed.

taliaga: In general, we have some checks and return some 400's when the request is really bad (see https://github.com/smartsdk/ngsi-timeseries-api/pull/250/commits/71dd4b1a6649e102ef79173c59426c240c637004 making this a bit better hopefully). For sure we don't cover all cases. It's easier to see when specific values are wrong, but if you have examples of combinations in mind, it'll be easier to check those specifically.

chicco785: not sure :) because wrong combination can lead to screw up sql that's all. not something also to fix in this pr, it was just a check

PR #255, opened 2019-09-30
Handling of ql/v2 root endpoint

### LINKED ISSUE #217: Handle ql/v2 root endpoint
The root endpoint (https://ql/v2/) should not error out, maybe present the api with the different options available

iamarnavgarg:
Fixes #217 as per the approach discussed.

### DIFF
--- a/specification/quantumleap.yml
+++ b/specification/quantumleap.yml
@@ -3,7 +3,6 @@ info:
   title: "QuantumLeap API"
   version: "0.7.5"      # we'll keep it aligned with QL version
 host: "localhost:8668"  # it'll run in the same container, hence localhost.
-basePath: /v2
 produces:
   - text/plain
   - application/json
@@ -319,6 +318,17 @@ parameters:
 # PATHS: META
 ################################################################################
 paths:
+
+  /v2/:
+    get:
+      operationId: reporter.api.list_of_api
+      summary: "Returns the List of API's available in QuantumLeap."
+      tags:
+        - meta
+      responses:
+        200:
+          description: "Successful response."
+
   /version:
     get:
       operationId: reporter.version.version
@@ -339,7 +349,7 @@ paths:
                 "version": "0.0.1"
               }
 
-  /config:
+  /v2/config:
     post:
       operationId: reporter.reporter.config
       summary: "(To Be Implemented) Customize your persistance configuration to
@@ -361,7 +371,7 @@ paths:
           schema:
             type: string
 
-  /health:
+  /v2/health:
     get:
       operationId: reporter.health.get_health
       summary: "Returns the health-check status of QuantumLeap and the services
@@ -469,7 +479,7 @@ paths:
 ################################################################################
 # PATHS: INPUT
 ################################################################################
-  /notify:
+  /v2/notify:
     post:
       operationId: reporter.reporter.notify
       summary: "Notify QuantumLeap the arrival of a new NGSI notification."
@@ -492,7 +502,7 @@ paths:
         500:
           description: "Internal server error."
 
-  /subscribe:
+  /v2/subscribe:
     post:
       operationId: reporter.reporter.subscribe
       summary: "Subscribe QL to process Orion notifications of certain type."
@@ -630,7 +640,7 @@ paths:
 ################################################################################
 # PATHS: OUTPUT
 ################################################################################
-  /entities/{entityId}/attrs/{attrName}:
+  /v2/entities/{entityId}/attrs/{attrName}:
     get:
       operationId: reporter.query_1T1E1A.query_1T1E1A
       summary: "History of an attribute of a given entity instance."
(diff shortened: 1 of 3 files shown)


specification/quantumleap.yml
```diff
@@ -3,7 +3,6 @@ info:
   title: "QuantumLeap API"
   version: "0.7.5"      # we'll keep it aligned with QL version
 host: "localhost:8668"  # it'll run in the same container, hence localhost.
-basePath: /v2
```
c0c0n3: why get rid of the base path? without it, we'll have to be explicit like you've done below, i.e. we'll have to use

    /v2/
    /v2/config
    /v2/version
    ...

instead of 

    /
    /config
    /version
    ...

So that changing version number will require updating all paths...but I'm sure there must a good reason for getting rid of the base path?

src/reporter/api.py
```diff
@@ -0,0 +1,21 @@
+
+def list_of_api():
+    return {
```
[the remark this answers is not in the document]
c0c0n3: I also agree with @chicco785. Also, unless we can automatically generate this file from the API spec, paths are bound to get out of sync since it's easy to change a path in one place and forget to do it in the other. I'd rather avoid duplicating paths if possible :-)

specification/quantumleap.yml
```diff
@@ -319,6 +318,17 @@ parameters:
...
+    get:
+      operationId: reporter.api.list_of_api
+      summary: "Returns the List of API's available in QuantumLeap."
+      tags:
+        - meta
+      responses:
+        200:
+          description: "Successful response."
+
   /version:
```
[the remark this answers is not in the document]
iamarnavgarg: @chicco785. As per my understanding, I have changed the paths separately by appending /v2/ in the beginning and there is no need for other code changes. Could you confirm with my understanding ?

src/reporter/api.py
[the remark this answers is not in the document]
iamarnavgarg: @chicco785 , I understand your concern. But as per @taliaga comment, we considered to list down all the available API's. So, if you can please confirm which API needs to be added to proceed further?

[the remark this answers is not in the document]
chicco785: sorry, my message was not clear, i meant that we don't need to explode the apis:

```
{
'config_url': '/v2/config',
'health_url': '/v2/health',
'notify_url': '/v2/notify',
'subscriptions_url': '/v2/subscriptions',
'entities_url': '/v2/entities',
'types_url': '/v2/types',
'attributes_url': '/v2/attrs'
}
```

specification/quantumleap.yml
chicco785: @c0c0n3 i actually think @iamarnavgarg is right, in the sense, that '/version' '/config' and '/health' should not be under '/v2' because they are not related with NGSI

specification/quantumleap.yml
```diff
@@ -319,7 +318,18 @@ parameters:
...
+    get:
+      operationId: reporter.api.list_of_api
+      summary: "Returns the List of API's available in QuantumLeap."
+      tags:
+        - meta
+      responses:
+        200:
+          description: "Successful response."
+
+  /v2/version:
```
chicco785: as said, I think it is more clean to not have `version` under v2

specification/quantumleap.yml
```diff
@@ -339,7 +349,7 @@ paths:
                 "version": "0.0.1"
               }
 
-  /config:
+  /v2/config:
```
chicco785: as said, I think it is more clean to not have `config` under v2

specification/quantumleap.yml
```diff
@@ -361,7 +371,7 @@ paths:
           schema:
             type: string
 
-  /health:
+  /v2/health:
```
chicco785: as said, I think it is more clean to not have `health` under v2

iamarnavgarg:
Hi @chicco785 ,

The build has completed all its steps successfully, But still in the end it shows that build gets no output. Can you please suggest any remedies for the same?

image

chicco785:
it looks like something is broken in the ci pipeline, i assume this is either due to a change in the alpine docker image or in travis ci, we will look into that

chicco785:
@iamarnavgarg PR #260 should fix the issue. when @c0c0n3 reviews it and we merge it, you should please rebase your code on the master. will let you know when you can proceed

c0c0n3:
@iamarnavgarg PR #260 just got merged into master. If like @chicco785 said, you could please rebase your branch on master so we can merge your PR in too. Many thanks!!!

iamarnavgarg:
Hi @c0c0n3 I have rebased/updated the branch, request you to please merge the PR.
Thanks!

c0c0n3:
Hi @iamarnavgarg, your PR just got merged into master, thank you so much for working on this!

PR #257, opened 2019-10-08
Remove obsolete static documentation

jason-fox:
@chicco785 @taliaga 

Replace with forwarder from GH-pages to ReadTheDocs.

I just received an odd eMail about the old *deprecated* Crate-DB - turns out the user was referring to the static documentation available under the GH-Pages. This needs to be either deleted or replaced with a forwarder to the **current** documentation.

Personally I prefer the latter, as there may be some inbound links out there.


chicco785:
thx @jason-fox for spotting the issue

Issue #162, opened 2019-02-12
Make /v2/entities return available ids


taliaga:
It may be useful if `/v2/entities` returned at least the ids of the available entities to further query for.

pooja1pathak:
Hi @taliaga , I am working on this issue. Please assign it to me.

chicco785:
hi @pooja1pathak 
performance wise, rather than using the `query` function in the translator, it may be easier and more performing to create a specific function.

1. you can use pagination (offset and limit) to return the list of tables, e.g.
`SELECT DISTINCT "table_name" FROM "doc"."md_ets_metadata" LIMIT 100;`
2. then you can use the list of tables to build a union query with pagination:
`SELECT entity_id, entity_type, max (time_index) time FROM "mtekz"."etdevice" GROUP BY entity_id, entity_type 
UNION ALL SELECT entity_id, entity_type, max (time_index) FROM "mtekz"."etstreetlight" GROUP BY entity_id, entity_type
 limit 100;`
3. in the result, it would be nice to return also entity_type and max index (see example above).
4. it would be nice to be able to filter by entity_type and time index.

c0c0n3:
closed by #264.

PR #264, opened 2019-10-22
Make /v2/entities return available ids. Fix for issue #162

pooja1pathak:
Fix for https://github.com/smartsdk/ngsi-timeseries-api/issues/162


c0c0n3:
hi @pooja1pathak, nice first shot at it :-)
we need to start thinking about performance though---not just for this PR but QL in general. The translator query method sucks into memory the whole result set, so filtering entity ids like you do in the reporter is less than optimal in general. Think of a prod scenario where there are 100 entity tables in Crate, each with 1,000,000 records. If I read the code correctly, a

    GET /v2/entities?limit=1000000000    # i.e. limit = 1 billion!

would cause QL to do something along the lines of:

    entities = []
    for each entity table T
        entities += SELECT * FROM T limit 1000000000
   
    # so we issued 100 queries and 
    # now we have 100 * 1000000 = 100M entities in memory 

At this point your code kicks in:

    ids = [ e.id for e in entities ]

So if a nasty client specifies e.g. `limit=1000000000`, QL would eventually run out of memory and die of a horrible death, hammering Crate in the process.

Obviously the problem isn't quite your code but rather the way the translator works. A solution based on streaming would be better here since we could do this in constant memory, but that would take quite a bit of work in the translator. @chicco785 sketched out a stop gap solution in #162, did you consider that?

chicco785:
@pooja1pathak thanks, can you rebase the code? also, to gain more in performance, we will need to introduce a "distinct" in the select. which means adding support for this in the `query` method.

pooja1pathak:
Hi @chicco785 , as per comment in https://github.com/smartsdk/ngsi-timeseries-api/issues/162#issuecomment-544926485 and https://github.com/smartsdk/ngsi-timeseries-api/pull/264#issuecomment-547385234 , I thought of writing a new method for ```/v2/entities```.

I have few queries:
1. Do I need to add ```entityType``` and ```TimeIndex``` as per https://github.com/smartsdk/ngsi-timeseries-api/issues/162#issuecomment-544926485.
2. How will I handle the scenario as descibed in https://github.com/smartsdk/ngsi-timeseries-api/pull/264#issuecomment-547385234
3. Do I need to write new method for ```/v2/entities``` or use the ```query``` method? If no, do I need to add ```distinct``` in select of the ```query``` method, as all the other queries are using this method. Will there be no Impact on them?

c0c0n3:
Hi @pooja1pathak  

> writing a new method for /v2/entities.

cool bananas!

> add entityType and TimeIndex as per #162 (comment).

that'd be great!

> How will I handle the scenario as descibed in #264 (comment)

Moving the filtering logic in the SQL query would already go a long way. Of course we still have the potential hot spot of loading the whole result set into memory, but that's a problem with QL everywhere and like I said a better approach would be to **stream** the records---think **constant** space---but that's too much of a change so do **not** worry about that. I'd simply add a cap to the limit param you get in for now, e.g.

    limit = max(limit_query_param, 10000)

> write new method for /v2/entities or use the query method?

rather use a new method? The query method is too hard to change I reckon and also the SQL query you'd most likely want to build isn't compatible with the ones built by the query method if my memory serves me well...

> add distinct in select of the query method, as all the other queries are using this method.

You hit the nail on the head! Like I said, I reckon the best option is to leave the query method be and write a new one to build the query outlined in the [comment to #162](https://github.com/smartsdk/ngsi-timeseries-api/issues/162#issuecomment-544926485)

pooja1pathak:
Hi @c0c0n3 ,

I am able to get response for one entity as:

```
[
  {
    "id": "Room1",
    "index": [
      "2019-10-30T16:39:37.902"
    ],
    "type": "RoomTest"
  }
]
```
Is this what we are expecting?

Also, I am thinking of adding ```fromDate```, ```toDate```, ```type```, ```offset``` and ```limit``` in query parameters. These values are also allowed in Orion. Is it ok to add only these parameters?

c0c0n3:
hi @pooja1pathak 

>  am able to get response for one entity

That's grand!

> Is this what we are expecting?

Spot on!

> Is it ok to add only these parameters?

Perfect. `offset` and `limit` we definitely need to have---think pagination of big datasets---the others would be highly desirable to keep the query interface similar to Orion's, thanks for pointing this out 👍

pooja1pathak:
Hi @c0c0n3 , just wanted to know about test-cases. I have created a test file but don't know how to run this test file for ql. Please guide me how to run test files.

c0c0n3:
Hi @pooja1pathak :-)

Test files go in the `tests` directories of the module you're testing, e.g. `src/reporter/tests/`. Our test framework is [pytest](https://docs.pytest.org/). Each test file name should start with a `test_`  and so should test function names within the file. Anyhoo, I reckon you figured that out all by yourself already a long time ago---well at least so it seems, judging by the test file in your latest commit: `src/reporter/tests/test_NTNE.py` :-)

Sorry for digressing. On to your question. To run test cases, you'll have to prep your shell env, which you do by running the `setup_dev_env.sh` script in the repo root dir:

    $ cd ngsi-timeseries-api/
    $ sh setup_dev_env.sh 

Then you can run **every** test case known to man with:

    $ sh run_tests.sh

This is painstakingly slow though since most tests are integration or end to end tests (rather than unit!) which run in a containerised environment through docker compose. So you'd typically only do that before a commit to double check the new commit doesn't wreak havoc. To speed up development a notch, you could try running just the test cases in your module, e.g. those in `src/reporter/tests/`. Each test dir comes with a `run_tests.sh` script that's called by the main `run_tests.sh` in the root dir. These scripts build the QL container image, bring up the containerised env as spec'd by `docker-compose.yml` (in the same test dir) and finally start a `pytest` session to run the module's test cases. Long story short: to run e.g. reporter tests, just:

    $ cd src/reporter/tests/
    $ sh run_tests.sh

If you have any unit tests, you can run them separately using your IDE or `pytest` directly from the shell. This way you won't have to twiddle your thumbs waiting on integration and e2e tests to cross the finishing line...

Hope that helps!

specification/quantumleap.yml
```diff
@@ -640,6 +652,49 @@ paths:
 ################################################################################
 # PATHS: OUTPUT
 ################################################################################
+  /v2/entities:
+    get:
+      operationId: reporter.query_NTNE.query_NTNE
+      summary: "List of all the entityId"
```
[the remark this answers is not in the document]
c0c0n3: good point. I haven't thought about it, but all things being equal I'd say yes?

chicco785:
> Hi @c0c0n3 ,
> 
> I am able to get response for one entity as:
> 
> ```
> [
>   {
>     "id": "Room1",
>     "index": [
>       "2019-10-30T16:39:37.902"
>     ],
>     "type": "RoomTest"
>   }
> ]
> ```
> 
> Is this what we are expecting?
> 
> Also, I am thinking of adding `fromDate`, `toDate`, `type`, `offset` and `limit` in query parameters. These values are also allowed in Orion. Is it ok to add only these parameters?

what would be the logic of `fromDate`, `toDate` ? i don't see need for these two parameters

totally fine `type`, indeed!!!

c0c0n3:
> what would be the logic of fromDate, toDate ?

Technically the set of entities changes over time, so it's legit to ask e.g. what was the set of entities a week ago. Not sure about actual use cases for this though...

specification/quantumleap.yml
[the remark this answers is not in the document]
pooja1pathak: Hi @taliaga , currently I am getting sorted response wrt time-index. Is it ok or I need to sort them on any other basis.

src/reporter/tests/test_NTNE.py
```diff
@@ -0,0 +1,41 @@
...
+def query_url():
+    url = "{qlUrl}/entities"
+
+    return url.format(
+        qlUrl=QL_URL
+    )
+
+@pytest.fixture()
+def reporter_dataset(translator):
+    insert_test_data(translator, [entity_type], n_entities=1, index_size=30)
```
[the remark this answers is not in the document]
pooja1pathak: Yes I have updated the test cases, I will update it in PR.

pooja1pathak:
> what would be the logic of `fromDate`, `toDate` ? i don't see need for these two parameters

My thinking was to get the list of entities that was created between ```fromDate``` ```toDate```. Please find below example:

I have created an entity in ql via /v2/notify at 09:30:25 and updated the same entity at 09:36:38. Now when I query ```/v2/entities``` for ```fromDate=2019-12-03T09:28:27``` ```toDate=2019-12-03T09:32:27```, it is showing below output:
```
[
  {
    "id": "Room2",
    "index": [
      "2019-12-03T09:30:25.698"
    ],
    "type": "RoomTest"
  }
]
```

Again if I query ```/v2/entities``` for ```fromDate=2019-12-03T09:28:27``` ```toDate=2019-12-03T09:50:27```, it is showing below output:

```
[
  {
    "id": "Room2",
    "index": [
      "2019-12-03T09:36:38.615"
    ],
    "type": "RoomTest"
  }
]
```
The idea here is to show all the entities which have at lease one value in the time frame provided. 

Please confirm is it ok or i need to update code.

specification/quantumleap.yml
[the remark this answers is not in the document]
c0c0n3: Sounds good to me. I reckon ordering by time index is probably what most peeps would expect?

c0c0n3:
@pooja1pathak thanks for explaining :-)

Like I said earlier, it looks like a reasonable thing to do to me too. Let's dot the i's and cross the t's though. Using the `fromDate` and `toDate` params we get a data slice `S` for each table `T` like this: 

1. `from ≤ to`. Then `S = { r ∈ T | r.time_index ∈ [from, to] }`.
2. `from` not given. Then `S = { r ∈ T | r.time_index ≤ to }`.
3. `to` not given. Then `S = { r ∈ T |  from ≤ r.time_index }`.
4. no params given. Then `S = T`, i.e. if the request has no time params, then we consider the **whole** table.

**For each** table `T` in QL, the query should pick out the "latest" entities in the corresponding slice `S` and return their union. Let's add a bit more detail, mainly for my own benefit so I'll be able to tell how this was supposed to work a month from now :-)

Within each `S`, we can group records by `entity_id`. Technically, we're defining an equivalence relation on `S`:

    r1 ~ r2  iff r1.entity_id = r2.entity_id

and each group ∈ `S/~ = { equivalence classes }`. There's a function that picks out the "latest" record in each group (equivalence class):

    latest : S/~  --->  T
    latest(g)  :=  r ∈ g  s.t.  x ∈ g ==> x.time_index ≤ r.time_index

Then we can associate to each table `T` in QL the set of latest entities in the corresponding table slice `S`:

    L(T) = Img(latest) = { latest(g) | g ∈ S/~ }

The query should build the **union** of such `L(T)`, i.e. 

    R = L(RoomTest) ∪ L(MyDevice) ∪ L(YourDevice) ∪ ...

and then return the set:

    { (r.entity_id, r.time_index, r.entity_type) | r ∈ R }

Hope that makes sense to you too?

chicco785:
not sure math explanation is clear to  @pooja1pathak
<https://github.com/pooja1pathak> :)

that is more or less the translation in term of sql query:

SELECT t1.entity_id, t1.entity_type, t1.time_index FROM
"serviceSchema"."entityTable" as t1 join  ( select entity_id,
max(time_index) time_index from "serviceSchema"."entityTable" WHERE
time_index > $from AND time_index < $to group by entity_id OFFSET $offset
LIMIT $limit) t2 on t1.entity_id = t2.entity_id and t1.time_index =
t2.time_index WHERE time_index > $from AND time_index < $to;

chicco785:
probably you don't even need the join, sorry for the mess :)

select entity_id, entity_type, max(time_index) time_index from
"serviceSchema"."entityTable" WHERE time_index > $from AND time_index <
$to group by entity_id, entity_type LIMIT $limit OFFSET $offset;

c0c0n3:
cool, thanks :-) the last query looks good but it only covers case (1) above, i.e. when we have **both** `fromDate` and `toDate` but it can easily be used for all other cases if we preprocess the input params, e.g.

    from = fromDate if fromDate else now
    to = toDate if toDate else 01-01-1970
    check from <= to ...

then your query template needs to be applied to all tables and then take the fat honking union of the result sets. the generated query should look similar to (pseudo-sql!!!):

    select entity_id, entity_type, max(time_index) time_index from
    "serviceSchema"."entityTable_1" 
    WHERE time_index >= $from AND time_index <= $to 
    group by entity_id, entity_type 
    
    UNION

    select entity_id, entity_type, max(time_index) time_index from
    "serviceSchema"."entityTable_2" 
    WHERE time_index >= $from AND time_index <= $to 
    group by entity_id, entity_type

    UNION
    ... "entityTable_3" ... 
    UNION
    ... "entityTable_n" ... 

    LIMIT $limit OFFSET $offset;

Please you guys double-check this is what we actually want :-)

pooja1pathak:
Hi @chicco785 @c0c0n3 , I have implemented similar type of query in my patch. Please check the updated PR and if it is OK, then merge the PR.

src/translators/crate.py
```diff
@@ -701,6 +701,75 @@ def query(self,
...
+                           )
+            else:
+                stmt += "select entity_id, entity_type, max(time_index) " \
+                           "from {tn} {where_clause}" \
+                           "group by entity_id, entity_type ".format(
+                              tn=tn,
+                              where_clause=where_clause
+                           )
+
+        op = stmt + "limit {limit} offset {offset}".format(
```
chicco785: as suggested, it may be a good idea ordering them by time index

src/reporter/tests/test_NTNE.py
```diff
@@ -0,0 +1,197 @@
...
+    # Query
+    query_params = {
+        'offset': 1
+    }
+    r = requests.get(query_url(), params=query_params)
+    assert r.status_code == 200, r.text
+
+    # Expect only last N
+    expected_values = list(range(1, n_days))
+    expected_type = 'Kitchen'
```
chicco785: https://travis-ci.org/smartsdk/ngsi-timeseries-api/builds/620582396#L1289 shows that you may get room actually. to be checked.

chicco785:
@pooja1pathak approach looks good, something is anyhow to be checked since test fails, see line: https://travis-ci.org/smartsdk/ngsi-timeseries-api/builds/620582396#L1289
most probably, what's to be checked is the actual test function.

src/reporter/tests/test_NTNE.py
pooja1pathak: Output is quite unpredictable for me. Sometime it shows Room and the sometime Kitchen. Don't know how to resolve this. Although all the test cases are passed now.

c0c0n3: > Output is quite unpredictable

I bet it's cos the result set isn't ordered by time. So you'll get the results in a different order every time you run the test. Can you check this? Thank you so much!

src/translators/crate.py
c0c0n3: yep, see also my comment above about failing test case.

src/translators/crate.py
```diff
@@ -701,6 +701,75 @@ def query(self,
...
+        offset = max(0, offset)
+        len_tn = 0
+        result = []
+        stmt = ''
+        for tn in table_names:
+            len_tn += 1
+            if len_tn != len(table_names):
+                stmt += "select entity_id, entity_type, max(time_index) " \
+                           "from {tn} {where_clause}" \
+                           "group by entity_id, entity_type " \
```
c0c0n3: Like I said in an earlier comment, I don't think grouping by `entity_type` is necessary since all rows should have the same value. This is a potential performance hot spot. If it's an easy thing to do, I'd just replace:

    select entity_id, entity_type, max(time_index)

with

    select entity_id, '<current entity type>' as entity_type, max(time_index)

where `current entity type` comes from the meta table you scan (if it's there, can't remember) or you could just

    (select entity_type from {tn} limit 1) as entity_type

Just an idea. But don't bother if this is a train smash for you. We're going to run a performance optimisation task at some point in time, we can easily tackle this and other QL issues then.

c0c0n3: Oh forgot to mention. The reason I said it could be a train smash, is that you'd have to verify my theory is actually correct and the query adjustment I suggested is better than what you have right now in terms of query execution plan, response time, memory usage, etc.

src/reporter/tests/test_NTNE.py
pooja1pathak: I found the cause of failed testcases that was because of the time of data inserted. Both Room and Kitchen value is inserted at the same time. To resolve this, I have updated the test cases by changing the value of time-index.

src/translators/crate.py
pooja1pathak: I have added code for order by time-index in ascending order as used in Orion.

src/reporter/tests/test_NTNE.py
c0c0n3: good catch!

src/translators/crate.py
c0c0n3: good stuff, thanks a stack!

pooja1pathak:
I have updated the PR as per the comments received. If it is ok then please merge the PR.

c0c0n3:
@pooja1pathak just merged in. thank you soooo much for your hard work and dedication, much appreciated!!!

PR #271, opened 2019-10-31
fix #269

### LINKED ISSUE #269: Travis tests fail
**Describe the bug**
It looks like #260 is coming back to haunt us :-o

**To Reproduce**
Look at Travis logs for #266, #268 

**Expected behavior**
Clean Travis test run.

**Environment (please complete the following information):**
 - Travis

**Additional context**
N/A

chicco785:


c0c0n3:
Commit 07418f4 implements a stopgap solution to #269 by pinning the Docker image base to Alpine `3.8`. I still don't know exactly which dependency caused this mayhem in Alpine `3.10`, but see comments to #269. So the good news is that we can move forward, the bad is that at some point we'll have to figure out how to upgrade to Alpine `3.10`.

Issue #209, opened 2019-06-13
don't store entries where all values are null


chicco785:
**Is your feature request related to a problem? Please describe.**

When you first register an iot device, a registration with empty values is sent to the context broker. as result a corresponding notification may be sent to quantumleap. such notification should not be used, given that no data will be in it.

Chandradeep-NEC:
@chicco785 @c0c0n3 @taliaga 
If no one is working on this issue, then lease assign us

daminichopra:
Hi @chicco785 , I am working on this issue and i want to confirm if payload contains both **"null"** and **some value** as Key value in payload for attributes in that case QL will filter out such entries and will not be stored in CrateDB.

chicco785:
@daminichopra the idea is to ignore only entities where all attribute values are null (except entityType and entityId).

daminichopra:
Hi @chicco785 , Please close this issue as it is Fixed. Thank you :)

c0c0n3:
cool, thanks for reminding us :-)
closed by #278

PR #278, opened 2019-11-20
Fix for Issue#209

daminichopra:
Changes done in reporter.py  for Issue https://github.com/smartsdk/ngsi-timeseries-api/issues/209


src/reporter/reporter.py
```diff
@@ -147,15 +147,26 @@ def notify():
...
     else:
         fiware_sp = None
-
-    # Send valid entities to translator
-    with translator_for(fiware_s) as trans:
-        trans.insert(payload, fiware_s, fiware_sp)
-
-    msg = 'Notification successfully processed'
+    for entity in payload:
+        attrs = list(iter_entity_attrs(entity))
```
c0c0n3: isn't this getting the attributes of the last entity in the payload? e.g. if

    payload = [ { a1{...}, a2{...} } , { a2{...}, a3{...} } ]

then `attrs = a2, a3` ? If I understand your algo correctly, this should cause a problem down below in

    value = payload[i][j]['value']

i.e. what's going to happen when evaluating `payload[0]['a3']['value']` ?
Perhaps I'm missing something?

src/reporter/reporter.py
```diff
@@ -147,15 +147,26 @@ def notify():
...
+    attrs.pop()
```
c0c0n3: why pop here? sorry I'm probably missing something...

src/reporter/reporter.py
```diff
@@ -147,15 +147,26 @@ def notify():
...
+    Flag = False
+    for i in range(length):
+        for j in attrs:
+            value = payload[i][j]['value']
+            if not value:
+                msg = 'Notification Not processed because of Null value'
+            else:
+                Flag = True
+                msg = 'Notification successfully processed'
+        if Flag:
```
c0c0n3: what if the payload contains three entities

    payload = e1, e2, e3

 `e1` and `e3` have values, `e2` does **not**. Isn't this sending the whole payload to the translator on the first loop iteration and then again on the third? Again sorry if there's something I'm missing, probably there is :-)

taliaga:
Hi @daminichopra !
If I may give a suggestion here, before working on `reporter.py`, how about starting with a test to cover the case mentioned in the issue (one entity with a valid type, id but just null values for all its attributes). A good candidate file for the test may be this one: https://github.com/smartsdk/ngsi-timeseries-api/blob/master/src/reporter/tests/test_incomplete_entities.py. Write the test, watch it fail and only then come and edit this file, until you make your test pass :)

src/reporter/reporter.py
daminichopra: Hi @c0c0n3 , I have used attrs.pop() to filter out time_index value from attrs.

src/reporter/reporter.py
daminichopra: @c0c0n3 , sorry i have missed that i have checked it for one entity. I am working on it.

src/reporter/reporter.py
c0c0n3: Ha! That guy! I totally forgot about that. But in general, we shouldn't assume `time_index` comes first... Anyhoo, all this shouldn't be an issue if put the new code into that `_filter_empty_entities` method I mentioned earlier and then call `_filter_empty_entities` **before** the `time_index` gets added---see my suggestion above.

src/reporter/reporter.py
c0c0n3: no worries, there's one gazillion things I miss too which is why we have code review :-)
thank you so much for working on this!

daminichopra:
Hi @c0c0n3 , i have Incorporated  your comments on PR, kindly review and merge the PR.

daminichopra:
Hi @c0c0n3 , i have incorporated review comments on PR. kindly review and close this PR. Thanks :)

src/reporter/reporter.py
```diff
@@ -104,7 +104,25 @@ def _validate_payload(payload):
             payload[attr].update({'value': None})
             log().warning(
                 'An entity update is missing value for attribute {}'.format(attr))
+  
 
+def _filter_empty_entities(payload):
+    log().info('Received payload: {}'.format(payload))
```
c0c0n3: isn't this redundant? there's already a log statement to output the whole entity array as we get it in the request:

https://github.com/smartsdk/ngsi-timeseries-api/pull/278/files#diff-f796da60daee45177c252e52fc78b4aaL120

src/reporter/reporter.py
```diff
@@ -104,7 +104,25 @@ def _validate_payload(payload):
...
+    attrs, length = list(iter_entity_attrs(payload)), (len(payload))
+    Flag = False
+    del_ids = []
+    for j in attrs:
+        key_list = list(payload[j].keys())
+        if 'value' not in payload[j]:
+            return payload
```
c0c0n3: um, not sure about this :-) what if the entity has an attribute with **no** value field as in `p` below:

    {
        'id': '299531',
        'type': 'AirQualityObserved',
        'p': { 'type': 'string' },
        'pm10': {'type': 'string', 'value': '', 'metadata': {}},
        'pm25': {'type': 'string', 'value': '', 'metadata': {}},
    }

Can you add a test case to check what happens? Thanks!

src/reporter/reporter.py
```diff
@@ -147,15 +168,26 @@ def notify():
         fiware_sp = request.headers.get('fiware-servicepath', None)
     else:
         fiware_sp = None
-
-    # Send valid entities to translator
-    with translator_for(fiware_s) as trans:
-        trans.insert(payload, fiware_s, fiware_sp)
-
-    msg = 'Notification successfully processed'
+    for entity in payload:
```
c0c0n3: Wasn't the whole `for` loop supposed to go? Not sure why it's still here. Am I missing something?

src/reporter/reporter.py
```diff
@@ -147,15 +168,26 @@ def notify():
         fiware_sp = request.headers.get('fiware-servicepath', None)
     else:
         fiware_sp = None
-
-    # Send valid entities to translator
-    with translator_for(fiware_s) as trans:
-        trans.insert(payload, fiware_s, fiware_sp)
-
-    msg = 'Notification successfully processed'
```
c0c0n3: I thought we should get this code block back in and zap the `for` loop below? I reckon there's something I'm missing here, oh deary deary :-)

src/reporter/tests/test_notify.py
```diff
@@ -313,3 +313,15 @@ def test_no_value_in_notification(notification):
...
+    notification['data'][0] = {
+        'id': '299531',
+        'type': 'AirQualityObserved',
+        'p': {'type': 'string', 'value': '', 'metadata': {}},
+        'pm10': {'type': 'string', 'value': '', 'metadata': {}},
+        'pm25': {'type': 'string', 'value': '', 'metadata': {}},
+    }
+    url = '{}'.format(notify_url)
+    r = requests.post(url, data=json.dumps(notification), headers=HEADERS_PUT)
+    assert r.status_code == 200
```
c0c0n3: cool, but how do we know the entity wasn't inserted in the DB?

src/reporter/reporter.py
daminichopra: Hie @c0c0n3 , I have checked and removed log statement from def notify().  :)

src/reporter/reporter.py
daminichopra: @c0c0n3 , I have checked it for the entity having attribute with no value field as in p, and test case is passing for this case too. Also, i have attached screenshot for the same.Please check
QL_TestCase
Capture

c0c0n3: > assert r.status_code == 200

yep, but the problem here is that getting back a `200` does **not** imply the record wasn't added to the DB. Unless I've missed something---which is entirely possible, old age, don't judge me!---the following could actually happen:

1. the reporter gets hit with a payload of e.g. `[ { 'id': '299531', 'type': 'AirQualityObserved', 'p': { 'type': 'string' } } ]`. So there's just one entity which contains only a `p` attribute with no `value` field.
2. that entity gets passed to ` _filter_empty_entities`.
3. the `if` on [line 116](https://github.com/smartsdk/ngsi-timeseries-api/pull/278/files/4dc1bd1b17632cb9100497ddae376bb59482e11b#diff-f796da60daee45177c252e52fc78b4aaR116) evaluates to true, so the entity is passed on to the translator.
4. the translator inserts the entity in the `etairqualityobserved` table successfully.
5. since there's no error, the reporter returns a `200` response to the client.

If you want to test this with an end-to-end/integration test, then you'd have to query the table to make sure the row hasn't been inserted. Just checking the HTTP response code won't cut it. Why not write a unit tests for ` _filter_empty_entities` though, isn't it easier? E.g.

    payload = { 'id': '299531', 'type': 'AirQualityObserved', 'p': { 'type': 'string' } }
    assert _filter_empty_entities(payload) is None

Just an idea.

src/reporter/reporter.py
c0c0n3: cool bananas!

src/reporter/tests/test_notify.py
daminichopra: Hi @c0c0n3 , i  have updated test_notify.py  In test_no_value_for_attributes() first post the entity check status code 200 and then it will check for the  entity wasn't inserted in the DB and check for status code 404. :)

src/reporter/reporter.py
daminichopra: @c0c0n3 , I have updated test case to check entity  wasn't inserted in the DB . If required i will check unit test too.

src/reporter/tests/test_notify.py
c0c0n3: awesome sauce!

c0c0n3: awesome sauce!

src/reporter/reporter.py
c0c0n3: Yea, I saw that, thanks a stack! But like I said earlier what I'm not sure about is what's going to happen if the entity looks like this:

    { 'id': '299531', 'type': 'AirQualityObserved', 'p': { 'type': 'string' } }

which is why I suggested adding a unit test for it. In fact, what I'm still struggling to get a grip on is the below filtering logic:

    for j in attrs:
        key_list = list(payload[j].keys())
        if 'value' not in payload[j]:
            return payload

My (quite possibly wrong!) interpretation is that if an entity has an attribute with no `value` field, then we insert that entity. Is that correct? If so what's the reasoning behind it?

daminichopra: @c0c0n3 ,
In  https://github.com/smartsdk/ngsi-timeseries-api/blob/da95f8dd85bca0fcc0a129aa37cd47956b560911/src/reporter/reporter.py#L104 
if it doesn't have string `value` it will update it to None . https://github.com/smartsdk/ngsi-timeseries-api/blob/257a27dcf3927a53005c4244bbee8955eaa785d1/src/reporter/reporter.py#L116
and not insert that entity. :)

src/reporter/reporter.py
daminichopra: @c0c0n3 , when we receive payload the it will validate it below the code.

src/reporter/reporter.py
c0c0n3: There might still be a little snag. What's going to happen if we have this entity:

    payload = {
        'id': '299531',
        'type': 'AirQualityObserved',
        'p': { 'type': 'string' },
        'pm10': {'type': 'string', 'value': '10', 'metadata': {}},
    }

If I'm right, then:

    _filter_empty_entities(payload) is None

whereas what we want is:

    _filter_empty_entities(payload) == payload

i.e. we should **not** filter the entity out since there's an attribute holding some data, `pm10` that is.  But here's a cheap fix, probably you can come up with something better:

    ...
    Flag = False
    for j in attrs:
        value = payload[j].get('value')
        if value:
            Flag = True
    if Flag:
    ...

If you could please add a test for the above `payload` and if it breaks see if fixing `_filter_empty_entities` the way I've suggested helps---or fix it in any other way you see fit :-)

src/reporter/reporter.py
c0c0n3: coolio, thanks!

src/reporter/reporter.py
daminichopra: @c0c0n3 , I have updated the code for value in entity and in def notify() it will check for entity update and if valid entity is there it will send it to translator further.
 https://github.com/smartsdk/ngsi-timeseries-api/blob/9e09c292bafa819ff6685b4c4d5db6676666bc70/src/reporter/reporter.py#L115
https://github.com/smartsdk/ngsi-timeseries-api/blob/9e09c292bafa819ff6685b4c4d5db6676666bc70/src/reporter/reporter.py#L160

Also, added test case in test_notify.py in which it will check for empty value  and for empty string 'value' in test_no_value_for_attributes(notification)
https://github.com/smartsdk/ngsi-timeseries-api/blob/b73ace5407e0a28d2c5f1aa35712ccd4b3a76534/src/reporter/tests/test_notify.py#L318

src/reporter/reporter.py
```diff
@@ -116,24 +131,19 @@ def notify():
                'content.', 400
 
     payload = request.json['data']
-
-    log().info('Received payload: {}'.format(payload))
-
+    
     # preprocess and validate each entity update
     for entity in payload:
```
c0c0n3: I suggest moving `_filter_empty_entities` here. That is, it should be the first thing we do:

    res_entity = []
    for entity in payload:
        entity = _filter_empty_entities(entity)
        if entity is None:
            continue
       res_entity.append(e) 
    ...

This way we avoid having to use `pop` in `_filter_empty_entities` which is kinda fragile---how do you know the first iterated attribute will always be the time index?! Also as a nice side effect we avoid looping through the array twice.

src/reporter/reporter.py
```diff
@@ -104,7 +104,22 @@ def _validate_payload(payload):
             payload[attr].update({'value': None})
             log().warning(
                 'An entity update is missing value for attribute {}'.format(attr))
+  
 
+def _filter_empty_entities(payload):
+    log().info('Received payload: {}'.format(payload))
+    attrs = list(iter_entity_attrs(payload))
+    Flag = False
+    attrs.pop()
```
c0c0n3: Didn't we decide to get rid of `pop` here? It isn't needed if `_filter_empty_entities` gets called before we add the time index attribute to the entity, see my comment below :-)

src/reporter/reporter.py
```diff
@@ -147,7 +157,15 @@ def notify():
...
     else:
         fiware_sp = None
-
+    res_entity = []
+    e = None
+    for entity in payload:
+        # Validate entity update
+        e = _filter_empty_entities(entity)
+        if e is not None:
+            res_entity.append(e)
```
c0c0n3: If you could move this logic up at as I suggested in my other comment, it'd be great!

src/reporter/reporter.py
c0c0n3: That's great! Like I said though, I'd feel much more confident about my ability to refactor in the future if we also had a test to cover what's going to happen if the entity has both valid and empty attributes as in:

    payload = {
        'id': '299531',
        'type': 'AirQualityObserved',
        'p': { 'type': 'string' },
        'pm10': {'type': 'string', 'value': '10', 'metadata': {}},
    }

c0c0n3:
hi @daminichopra, I've just added a couple of comments about possible improvements---see above review. If you have time to review/validate/or amend my suggestions and then provide an implementation it'd be fantabulous. On the other hand, if there isn't more time you can dedicate to this PR, just let me know. I think as it stands now it's pretty good anyway so we might just as well merge it in if you agree.

daminichopra:
Hie @c0c0n3 , I have updated test case for valid attribute and empty attributes as well. Please close this PR and raise new issue for remaining comments. :)

src/reporter/tests/test_notify.py
```diff
@@ -313,3 +313,50 @@ def test_no_value_in_notification(notification):
...
+        'p': {'type': 'string'},
+        'pm10': {'type': 'string', 'value': '10', 'metadata': {}},
+    }
+    url = '{}'.format(notify_url)
+    get_url = "{}/entities/299531/attrs/p/value".format(QL_URL)
+    url_new = '{}'.format(get_url)
+    r = requests.post(url, data=json.dumps(notification), headers=HEADERS_PUT)
+    assert r.status_code == 200
+    res_get = requests.get(url_new, headers=HEADERS_PUT)
+    assert res_get.status_code == 404
```
c0c0n3: Do we get a `404` here cos the attribute isn't there or the whole entity is **not** in the DB? Off the top of my head I can't tell what QL does if you query an attribute of an entity that's **not** in the DB. So could it be we get a `404` because entity `299531` wasn't inserted? Can you please check that?

What we want to check with this test case is that the entity gets inserted with a `pm10` attribute value of `10`.  So I'd just check that too, e.g.

    pm10_url = "{}/entities/299531/attrs/pm10/value".format(QL_URL)
    res_get = requests.get(pm10_url, headers=HEADERS_PUT)
    assert res_get.status_code == 200
    assert get_value(res_get) == '10'

`get_value` is a made up method, can't remember how to extract the returned value from the response :-0

c0c0n3:
Hi @daminichopra thank you sooo much for this.

> raise new issue for remaining [comments]

Excellent idea! I'll do that. There's just a couple of things still hanging though as far as I can tell.

1. Your branch needs rebasing since it's behind master GH won't let me merge it.
2. I have question about the test case you added, see my comment [here](https://github.com/smartsdk/ngsi-timeseries-api/pull/278/files#r359756679). Sorry if I sound pedantic, but we need to be sure this test case proves the entity gets inserted. This is critical for us since it could break some of our deployments...

Thanks a stack!!!

src/reporter/reporter.py
daminichopra: Hie @c0c0n3 , I have updated TestCase for Both valid entity and empty value string.
https://github.com/smartsdk/ngsi-timeseries-api/blob/6ff27a7aae6b50cc423bda4acc5723e5c9782639/src/reporter/tests/test_notify.py#L363
https://github.com/smartsdk/ngsi-timeseries-api/blob/6ff27a7aae6b50cc423bda4acc5723e5c9782639/src/reporter/tests/test_notify.py#L365

In this it will check response to validate that the entity gets inserted or not using method res_get.json().

src/reporter/tests/test_notify.py
daminichopra: Hie @c0c0n3 , I have updated TestCase for Both valid entity and empty value string.
https://github.com/smartsdk/ngsi-timeseries-api/blob/6ff27a7aae6b50cc423bda4acc5723e5c9782639/src/reporter/tests/test_notify.py#L363
https://github.com/smartsdk/ngsi-timeseries-api/blob/6ff27a7aae6b50cc423bda4acc5723e5c9782639/src/reporter/tests/test_notify.py#L365

In this it will check response to validate that the entity gets inserted with a pm10 attribute value of 10 or not using method res_get.json().

c0c0n3: Good stuff! Thanks a stack!

c0c0n3:
@daminichopra I reckon this PR is ready for merging, if you could please **rebase onto master**. Thanks so much!

c0c0n3:
@daminichopra can you have a look at the failing tests? Perhaps something in your latest commit? One thing I could pick up is [this line](https://github.com/smartsdk/ngsi-timeseries-api/pull/278/files#diff-f796da60daee45177c252e52fc78b4aaR115) in `_filter_empty_entities`

    value = payload[j]['value']

should probably be

    value = payload[j].get('value')

as we [talked about](https://github.com/smartsdk/ngsi-timeseries-api/pull/278#discussion_r357269247) a while back. Not sure this is what makes the tests unhappy though, like I said if you could please have a look, many thanks!

c0c0n3:
@daminichopra it looks like this PR is ready for merging, if you give me the green light I'll happily squash it into master.

We should also reimplement properly the tests in `test_incomplete_entities.py` that you made pass by checking there's no attrs. These tests were meant to check what happens to attrs with no value, so after your change we'll need to update the tests to use an entity with two attrs instead of one---the first attr will have to have a value, the other will be the `x` attr as it was in the tests before your change. 

Since this is just donkey work I'd rather merge your **valuable** contribution and then do the rest separately along with the few minor improvements we discussed earlier. Like you suggested I'll open a new issue for the boring little bits :-)

daminichopra:
Hi @c0c0n3 , Please close this PR if it looks good. Also, Please create a new issue for other minor improvments. Thank you :)

c0c0n3:
@daminichopra merged :-) kudos to you!

Issue #225, opened 2019-07-17
Support use of " wrapping fromDate and toDate


taliaga:
**Describe the bug**

Parameters dateFrom and dateTo of several API methods require a date in ISO format. If such paramter goes as "YYYY-mm-ddThh:mm:ss" instead of just YYYY-mm-ddThh:mm:ss (i.e. with wrapping "") QL does not recognize the dates.

pooja1pathak:
Hi @taliaga @c0c0n3 , I want to contribute for this issue. Please assign it to me.

c0c0n3:
@pooja1pathak just assigned to you, thanks :-)))

pooja1pathak:
@c0c0n3 , as PR #285 is merged, this issue can be closed.

c0c0n3:
@pooja1pathak whoops, thanks for reminding me!

closed by PR #285

PR #285, opened 2020-01-27
Fix for issue #225

pooja1pathak:
Fix for issue #225
Updated code to accept dateFrom and dateTo of API methods with quotes("").

### DIFF
--- a/src/reporter/delete.py
+++ b/src/reporter/delete.py
@@ -35,11 +35,6 @@ def delete_entity(entity_id, type_=None, from_date=None, to_date=None):
 def delete_entities(entity_type, from_date=None, to_date=None):
     fiware_s = request.headers.get('fiware-service', None)
     fiware_sp = request.headers.get('fiware-servicepath', None)
-    
-    if from_date:
-        from_date = from_date.strip('\"')
-    if to_date:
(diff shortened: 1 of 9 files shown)


c0c0n3:
@pooja1pathak this is great thanks, except...most tests are failing. It looks like there's issues building the QL image, if you could please look into that :-)

c0c0n3:
Also I'd like to propose an improvement. Can we move the strip stuff into a `parse_date` function which takes the string, strips any extra `"`, then attempts to convert to an actual Python date. Accordingly the translator should operate on date objects rather than strings. This way we avoid further grief if e.g. the date string is in a format we don't support. Also if in the future we decide to support more than one date format, we can just make the change in a single place, i.e. in `parse_date`.

This is just a suggestion. I'm actually already very happy with your submission. But just in case you have the time to do it...you know.

pooja1pathak:
Hi @c0c0n3 , Thanks for the review :-)

I have updated the code as per your suggestion. Please review the updated PR.

c0c0n3:
@pooja1pathak should I go ahead and merge into master?

pooja1pathak:
Yes you can merge this PR :-)

Issue #292, opened 2020-02-14
Fix geocoding tests


c0c0n3:
**Describe the bug**
Lately we've been getting no love from OSM.  Geocoding tests have been failing with e.g.

    Status code 403 from https://nominatim.openstreetmap.org/search: ERROR - 
    403 Client Error: Forbidden for url: https://nominatim.openstreetmap.org/search?q=IJzerlaan+18+%2C+Antwerpen+%2C+BE&format=jsonv2&addressdetails=1&limit=10

It doesn't look like a transient error (it's a `403`!) so perhaps the OSM API changed and we need tweaking our client?

**To Reproduce**
See e.g. PR #291's [build log](https://travis-ci.org/smartsdk/ngsi-timeseries-api/builds/650362025?utm_source=github_status&utm_medium=notification).

**Expected behavior**
Well, tests should pass?

**Environment (please complete the following information):**
 - Travis

**Additional context**
We've seen something similar in the past, see #105 and PR #144.

c0c0n3:
It look like we're *not* getting the above error anymore, but some others related to coordinates: see [build log](https://travis-ci.org/github/smartsdk/ngsi-timeseries-api/builds/669184574?utm_source=github_status&utm_medium=notification) of PR #294.

As a stopgap solution, PR #299 disabled the failing tests, see [diff](https://github.com/smartsdk/ngsi-timeseries-api/pull/299/files?file-filters%5B%5D=.py&file-filters%5B%5D=.yml#diff-2f5989997bcc89ce952bcca5ef20ed34)

PR #299, opened 2020-04-03
introduce support for log level configuration

### LINKED ISSUE #160: Log Level
How can I change log level?

chicco785:
fix #160 and introduce multiple level of logs

* differentia log levels
* add `LOGLEVEL` env variable
* document env variables

### DIFF
--- a/src/translators/crate.py
+++ b/src/translators/crate.py
@@ -63,9 +63,9 @@ class CrateTranslator(base_translator.BaseTranslator):
         url = "{}:{}".format(self.host, self.port)
         self.conn = client.connect([url], error_trace=True)
         self.cursor = self.conn.cursor()
-        #IMPORTANT this reduce queries to crate, but it also means
-        #that you need to redeploy ql in case of major crate update
-        #TODO document this
+        # IMPORTANT this reduce queries to crate, but it also means
+        # that you need to redeploy ql in case of major crate update
+        # TODO document this
         self.db_version = self.get_db_version()
 
     def dispose(self):
--- a/src/translators/crate.py
+++ b/src/translators/crate.py
@@ -63,9 +63,10 @@ class CrateTranslator(base_translator.BaseTranslator):
         url = "{}:{}".format(self.host, self.port)
         self.conn = client.connect([url], error_trace=True)
         self.cursor = self.conn.cursor()
-        # IMPORTANT this reduce queries to crate, but it also means
-        # that you need to redeploy ql in case of major crate update
-        # TODO document this
+        # TODO this reduce queries to crate,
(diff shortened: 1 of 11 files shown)


src/translators/crate.py
```diff
@@ -63,7 +63,10 @@ def setup(self):
         url = "{}:{}".format(self.host, self.port)
         self.conn = client.connect([url], error_trace=True)
         self.cursor = self.conn.cursor()
-
+        # IMPORTANT this reduce queries to crate, but it also means
+        # that you need to redeploy ql in case of major crate update
```
c0c0n3, line 67: Are you saying this because of global state? There's no global state here, you'll get a fresh translator instance each time the notify endpoint gets hit:

* https://github.com/smartsdk/ngsi-timeseries-api/blob/master/src/reporter/reporter.py#L172
* https://github.com/smartsdk/ngsi-timeseries-api/blob/master/src/translators/factory.py#L28
* https://github.com/smartsdk/ngsi-timeseries-api/blob/master/src/translators/crate.py#L1029
* https://github.com/smartsdk/ngsi-timeseries-api/blob/master/src/translators/base_translator.py#L19

Notice the use of a [context manager](https://docs.python.org/3/library/contextlib.html#contextlib.contextmanager)

If that's what you were worrying about I suggest we get rid of this sentence "but it also means that you need to redeploy ql in case of major crate update" since it's going to be confusing.

chicco785: then we can remove the comment, still is a query that consume time and 99% with the same result, so if we have somewhere this a static value that is not refreshed every notification, may be a better option.

chicco785: I updated the TODO text

chicco785:
>1 . I'm assuming you verified manually that the log level gets picked up from the env and then you can see log lines being output at that level or higher but not below. E.g. if env log level = info, then you should not see any debug lines in the logs.

Yes, have verified that, not sure we can develop a test though. It may be quite a pain.

>2. Have a look at the other comment in the review---trifling point :-)

Did that

>3. I added your TODOs about failing geocoding tests to the list of things to look into for #292, so we can deal with that separately, hope that's okay.

Top!

src/translators/crate.py
c0c0n3: yep, excellent point, I opened a separate issue just for that: #302

c0c0n3:
> not sure we can develop a test though. It may be quite a pain.

totally. visual inspection is more than enough for me.

Thanks for the PR, awesome work. I'm going to merge if there's nothing else you'd like to add.

chicco785:
yes please fill free to merge

Issue #256, opened 2019-10-01
CrateDB 4 Upgrade


c0c0n3:
### The problem
With the CrateDB 4.x series out since June 2019, we should consider upgrading at some point. 
Currently we support version `3.3` and it doesn't look like that upgrading to `4.x` is going to be a walk in the park for us. In fact, some of the SQL statements we generate rely on features that were dropped in Crate `v4` (see e.g. #254) and it'd be a good idea to review the rest of the SQL we generate too. Also, we might have to migrate prod tables since according to the Crate docs:

> Tables that were created prior CrateDB 3.x will not function with 4.x and must be recreated before moving to 4.x.x.

Do we have any prod tables that we created with Crate `2`? Did the upgrade to `v3` automatically changed their format so that they'll be compatible with `v4`? Um, I doubt it...

### Proposed solution
Test each and every SQL statement we generate with Crate 4 and fix what breaks. Figure out if we need a data migration too. Then do the upgrade.

### Alternatives
If testing SQL statements turns out to be too time consuming, we should consider if this is a good time to re-engineer the translators (both Crate and Timescale) to use ASTs instead of string manipulation.

chicco785:
@c0c0n3 as first action, i would in the readme the version supported of crate and timescale

chicco785:
`ON CONFLICT (column_ident [, ...]) DO UPDATE SET { column_ident = expression` update may not be that painful. Yes we had some release using crate 2.x. But i am not sure anyone is still using it.

c0c0n3:
> not be that painful.

Yep, as long as it's the only thing to change, but we should test the rest too :-)

tstorek:
Hi,
is there any update on this? From my perspective crate 4.x is slighly easier adminstrate and to scale than v3.x. Especially from the documentation point of view it seems easier. But maybe this also goes together with #258 which would make quantumleap more independent from the database engine itself.

Cheers

c0c0n3:
Hi @tstorek,

No, we haven't managed to tackle this yet. We're understaffed at the moment, so this will probably have to wait until next year---of course, contributions are welcome! Re: InfluxDB, we have no plans to support it at the moment since it's missing some critical (to us!) features, see [this comment](https://github.com/smartsdk/ngsi-timeseries-api/issues/258#issuecomment-541308721) to #258.

Sorry to disappoint you, but like I said we'd really need an extra pair of hands to get this off the ground...

pinkerltm:
[...]

PR #300, opened 2020-04-04
support crate 4.x

chicco785:
fix #256 

@c0c0n3 three test are failing. 
* `test_NTNE_limit` probably because the query is actually wrong
* `test_NTNE_offset` may be actually a crate bug. i run the query manually and it seems that crate 4.x is not handling correctly `offset` when using it with an aggregate query (at least if in a union all select)
* `test_bc` well of course moving to crate 4.x it's not backward compatible with ql 0.5.1

### DIFF
--- a/src/translators/crate.py
+++ b/src/translators/crate.py
@@ -30,6 +30,8 @@ CRATE_ARRAY_STR = 'array(string)'
 NGSI_TO_CRATE = {
     "Array": CRATE_ARRAY_STR,  # TODO #36: Support numeric arrays
     "Boolean": 'boolean',
+# TODO since CRATEDB 4.0 timestamp is deprecated. Should be replaced with timestampz
+# This means that to maintain both version, we will need a different mechanism
     NGSI_ISO8601: 'timestamp',
     NGSI_DATETIME: 'timestamp',
     "Integer": 'long',
@@ -204,7 +206,7 @@ class CrateTranslator(base_translator.BaseTranslator):
                 msg = "Translating entity without TIME_INDEX. " \
                       "It should have been inserted by the 'Reporter'. {}"
                 warnings.warn(msg.format(e))
-                now_iso = datetime.now().isoformat(timespec='milliseconds')
+                now_iso = datetime.utcnow().isoformat(timespec='milliseconds')
                 e[self.TIME_INDEX_NAME] = now_iso
 
         # Define column types
@@ -231,7 +233,8 @@ class CrateTranslator(base_translator.BaseTranslator):
                 if isinstance(e[attr], dict) and 'type' in e[attr]:
                     attr_t = e[attr]['type']
                 else:
-                    # Won't guess the type if used did't specify the type.
+                    # Won't guess the type if user did't specify the type.
+                    # TODO Guess Type!
                     attr_t = NGSI_TEXT
 
                 col = self._ea2cn(attr)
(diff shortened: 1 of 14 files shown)


c0c0n3:
@chicco785 it looks like you've added these changes on top of PR #299. We should freeze this PR until we merge PR #299 into master, then rebase this PR onto master to avoid version control nightmares :-)

For the rest it looks like a very good start. What happens if you re-enable all reporter tests? Talking about this change:

    - pytest src/reporter/ --cov-report= --cov-config=.coveragerc --cov-append --cov=src/
    + pytest src/reporter/tests/test_NTNE.py --cov-report= --cov-config=.coveragerc --cov-append --cov=src/

chicco785:
yep, the idea was to rebase after merging the log one, locally i already re-enabled all report tests. i disable to focus on the troublesome ones

chicco785:
I reported the Union all "bug" in crate github https://github.com/crate/crate/issues/9854

c0c0n3:
>  idea was to rebase after merging the log one, locally i already re-enabled all report tests

good stuff. I see you've pushed the change, I can rebase this branch if you like or you can do that yourself if you still have something you haven't pushed...

> disable to focus on the troublesome ones

cool bananas, just wanted to make sure the change was intentional :-)

>  reported the Union all "bug" 

that's great, thanks!

chicco785:
@c0c0n3 
in general i assume that when you have lot of entities, offset is more important than order by, so at the time being for the crate bug, i removed order by.

i also improved the backward compatibility test, so we are able to test also database upgrade, and to test it across different combination of quantumleap and crate. see `.travis.yml`

chicco785:
https://travis-ci.org/github/smartsdk/ngsi-timeseries-api/jobs/671718649?utm_medium=notification&utm_source=github_status highlights that query behaviour between crate versions in inconsistent

c0c0n3:
> in general i assume that when you have lot of entities, offset is more important than order by, so at the time being for the crate bug, i removed order by.

That seems the most sensible thing to do at the moment. I opened #306 as a reminder to follow up on the Crate issue and possibly revert back to ordering the result set.

> i also improved the backward compatibility test

That's fantabulous, thanks!

> query behaviour between crate versions in inconsistent
> A theory could be that the order of result - in absence of order by - depends on the order of consolidation by crate (and not on the order of insertion).

Yes probably. Since those tables don't have a PK, perhaps we get an implicit one with an associated index and results get output in index traversal order. But that's just a wild guess!

c0c0n3:
Also, I'd update `test_NTNE_limit` to make it independent of result set ordering. What we want to test here is that the limit param is honoured. So the test should just check the returned tuple `(id, type, ix) ∈ { (Room0, Room, room_ix), (Kitchen0, Kitchen, kitchen_ix) }`.

chicco785:
@c0c0n3 all test are passing

src/translators/crate.py
```diff
@@ -204,7 +206,7 @@ def _insert_entities_of_type(self,
                 msg = "Translating entity without TIME_INDEX. " \
                       "It should have been inserted by the 'Reporter'. {}"
                 warnings.warn(msg.format(e))
-                now_iso = datetime.now().isoformat(timespec='milliseconds')
+                now_iso = datetime.utcnow().isoformat(timespec='milliseconds')
```
chicco785, line 209: @c0c0n3 this ensure that also when not timeindex is inserted, we use utc

c0c0n3: excellent thanks!

chicco785:
@c0c0n3 unless there is anything else, this PR should be ready to land

c0c0n3:
@chicco785 excellent, merged in!

Issue #273, opened 2019-11-02
Upgrade Docker image to Alpine 3.10


c0c0n3:
**Is your feature request related to a problem? Please describe.**
The Docker image uses Alpine as a base. QuantumLeap won't start in the container if the image gets built from Alpine `3.10`---see #269 for the details. PR #271 implemented a stopgap solution by pinning the Alpine version to `3.8`.

**Describe the solution you'd like**
We need to understand what goes wrong exactly with `3.10` and then upgrade the image accordingly. Keep in mind the base version should still be pinned though as done in #271 to up the likelihood of build reproducibility. We definitely don't want to wind up again in a situation where it's possible to build two slightly different images depending on the machine you build on, courtesy of loose image tagging as in `python:3.6-alpine` where the same tag points to an image that gets regularly updated---the horror!

**Describe alternatives you've considered**
Docker layers, image tagging and caching, silly build file syntax, Linux package managers...these are 20th century solutions to dependency management. Nix and NixOS have leaped in the 21st century already offering a way out of the mess with proper dependency management and programming language support. We should consider a Nix-based build system so we can track all deps properly and guarantee proper app isolation and build reproducibility. As an added benefit, we can create tiny Docker images, even smaller than what you can do with an Alpine base.

**Additional context**
#271, #270, #269, #315 ([this comment in particular](https://github.com/smartsdk/ngsi-timeseries-api/pull/315#issuecomment-644818232))

c0c0n3:
PR #315 upgraded to Alpine `3.12` and Python `3.8.3`.

chicco785:
is this still a valid issue?

Issue #270, opened 2019-10-31
Failing to run up docker containers via docker-compose


mtalhaf:
While trying to run a local instance of quantum leap with all the other necessary tools via the  docker-compose provided, the quantum leap container fails with an error of "Failed to add operation for POST /v2/config", while the final error being "connexion.exceptions.ResolverError: <ResolverError: Cannot resolve operationId "reporter.reporter.config"! Import error was "No module named 'pytest'">"

**To Reproduce**
1. Download repository as zip
2. Navigate to docker folder
3. Run `docker-compose -f docker-compose-dev.yml up -d`

**Expected behavior**
All tools should run properly to be able to test quantum leap.

**Environment (please complete the following information):**
 - OS: Mac High Sierra 10.13.6

Find attached the logs form Quantum leap while running the docker-compose without the `-d `command

[Quantum-leap-error.txt](https://github.com/smartsdk/ngsi-timeseries-api/files/3793943/Quantum-leap-error.txt)

Any help with this is appreciated.

c0c0n3:
Hi @mtalhaf, it looks like your problem could be that you didn't run `pipenv install` (installs Python env + deps) and `setup_dev_env.sh` (sets up vars you'll need to run docker compose files). Try this:
 
    $ cd wheres/you/extracted/ngsi-timeseries-api
    $ pipenv install
    $ sh setup_dev_env.sh
    $ cd dir/where/docker-compose/you/want/to/run/is
    $ docker-compose up -d

Let me know if that works for you :-)

chicco785:
i think the point is which image has been used, if it is the one from the
docker hub repo,
there is surely a problem, a local one was used, I believe andrea is correct

kernchen:
@chicco785, @c0c0n3,

Just writing, since I've been working on this with @mtalhaf together. We basically tried out QuantumLeap. I tried it out a bit earlier with this hash, that didn't have any issues:

sha256:d310ee076e91a60c857c9f49dee5524149d52b44afe763e9b8ac661f0f0de6bb

It's coming directly from Docker Hub. We've been using the same docker-compose-dev.yml file. While mine processed fine, it says "2 weeks ago". @mtalhaf is now having issues with latest.

Hope that helps.

kernchen:
Maybe as an addition, would you recommend using 0.7.5 instead of latest?

c0c0n3:
[...]

PR #315, opened 2020-04-18
WIP: Fix 314

chicco785:
- [x] Step 0 - identify reusable code and provide separation:
  * define super class for SQL translator
  * move replicated code to super class
  * keep only specific code in crate translator
- [ ] Step 1 - support more flexible db manager:
  * replace base driver with SQL Alchemy driver (at a basic level, without leveraging most advanced features)
- [x] Step 2 - apply changes to timescale:
  * update timescale translator to extend SQL translator and use SQL Alchemy 
  * remove duplicated code from timescale translator
- [ ] Step 3 - understand if we got some free food:
  * explore changes needed to support timescale queries


chicco785:
step 1 - may not be that simple see https://github.com/crate/crate-python/issues/273

chicco785:
given that, same issue may apply to postgis, for the time being i may skip step 1 and 2.1

Dockerfile
```diff
@@ -1,19 +1,17 @@
...
 RUN ln -s /usr/include/locale.h /usr/include/xlocale.h
 RUN pip install pipenv
 RUN mkdir -p /src/ngsi-timeseries-api
 COPY Pipfile /src/ngsi-timeseries-api/Pipfile
 COPY Pipfile.lock /src/ngsi-timeseries-api/Pipfile.lock
-RUN mkdir /install
-WORKDIR /install
 RUN cd /src/ngsi-timeseries-api && { pipenv lock -r > /requirements.txt; }
-RUN pip install --install-option="--prefix=/install" -r /requirements.txt
+RUN pip install -r /requirements.txt
```
chicco785, line 10: not sure why, but installation in the old way was not working anymore, now it's working, but as result the image is bigger :/

src/translators/sql_translator.py
```diff
@@ -0,0 +1,964 @@
...
+
+        geo_clause = self._get_geo_clause(geo_query)
+        if geo_clause:
+            clauses.append(geo_clause)
+
+        where_clause = "where " + "and ".join(clauses)
+        return where_clause
+
+    def _parse_date(self, date):
+        try:
```
chicco785, line 400: in this way we should have tackled sql injection for dates

Dockerfile
chicco785: see comment here

chicco785: @c0c0n3

c0c0n3: thx!

src/translators/sql_translator.py
c0c0n3: awesome

chicco785:
@c0c0n3 fyi test failing should be mostly due to the test refactoring that was not completed of step3, where using ENV variables I am dynamically setting the usage of crate or timescale

c0c0n3:
that's cool, this is work in progress but thanks for the heads-up! we should fix them before merging into master though :-)

chicco785:
most probably if you split commit 2 from commit 3 as a first step of the
merge, not much may be needed.

c0c0n3:
@chicco785 so I think I've managed to fix up all the tests. Haven't managed to test w/ Timescale yet, but let's see what Travis has to say about it before I spend more time on this. Speaking of time

* a full test run takes about 13 mins now (on my laptop), down from 34 mins. still not good but at least acceptable, so we won't get totally stuck waiting for test feedback when developing locally.
* it took forever to get this right, mostly because the test code is **very** hard to maintain. the small change you made in `insert_test_data` caused a ripple effect---tests took almost double the time to run and basically all reporter tests needed tweaking. quite a steep price to pay for such a small change...
 
So we should put some thought into refactoring. I did that for ntne1a (see 0aa203f and 6cc1cd4) but then as we agreed I left the rest behind and patched all the other reporter tests---which are still a mess. Anyways, at least we should be able to press on now---famous last words :-)

chicco785:
> @chicco785 so I think I've managed to fix up all the tests. Haven't managed to test w/ Timescale yet, but let's see what Travis has to say about it before I spend more time on this. Speaking of time
> 
> * a full test run takes about 13 mins now (on my laptop), down from 34 mins. still not good but at least acceptable, so we won't get totally stuck waiting for test feedback when developing locally.
> * it took forever to get this right, mostly because the test code is **very** hard to maintain. the small change you made in `insert_test_data` caused a ripple effect---tests took almost double the time to run and basically all reporter tests needed tweaking. quite a steep price to pay for such a small change...
> 
> So we should put some thought into refactoring. I did that for ntne1a (see [0aa203f](https://github.com/smartsdk/ngsi-timeseries-api/commit/0aa203f6725d4cff83d75d5c2a8b6427f1f7b17c) and [6cc1cd4](https://github.com/smartsdk/ngsi-timeseries-api/commit/6cc1cd41afacf3f74949d7aa33ac138e4f73d747)) but then as we agreed I left the rest behind and patched all the other reporter tests---which are still a mess. Anyways, at least we should be able to press on now---famous last words :-)

it seems that build is still broken not sure though why

c0c0n3:
>  build is still broken not sure though why

yea me neither. ive done some digging and here's what I've got so far:

* after setting up my **local** test env **from scratch**, all tests pass with Crate 4.1.4
* but some of them are broken on Travis---w/ the **same** Crate version
* I've managed to reproduce some of the Travis failures locally with Crate 3.3.5
* it looks like most errors have to do with time format---UTC timezone, a baffling mystery, well unless there's some issues with Crate

will keep on digging...

chicco785:
> > build is still broken not sure though why
> 
> yea me neither. ive done some digging and here's what I've got so far:
> 
> * after setting up my **local** test env **from scratch**, all tests pass with Crate 4.1.4
> * but some of them are broken on Travis---w/ the **same** Crate version
> * I've managed to reproduce some of the Travis failures locally with Crate 3.3.5
> * it looks like most errors have to do with time format---UTC timezone, a baffling mystery, well unless there's some issues with Crate
> 
> will keep on digging...

not sure what's root source, but if you check the travis logs up to step-2 the issue was not there :/ https://travis-ci.org/github/smartsdk/ngsi-timeseries-api/jobs/680120440 

there also 3.3.5 was tested, so it's something that was changed after the commit https://github.com/smartsdk/ngsi-timeseries-api/commit/85d83bcdba4ac873e0dbae9662acd095527fbde8

chicco785:
so also up to step-3 the issue was not there:
https://travis-ci.org/github/smartsdk/ngsi-timeseries-api/jobs/691676310#L1011

not having intermediate builds of your changes, not sure in which exact point / which change triggered it

c0c0n3:
> check the travis logs up to step-2 the issue was not there

you've got me lost :-) the travis logs say step 2 was already broken? i.e. it looks like [build #1051](https://travis-ci.org/github/smartsdk/ngsi-timeseries-api/builds/680120439) for commit 85d83bc was broken already? Also it looks like it what broke it was related to time format:

https://travis-ci.org/github/smartsdk/ngsi-timeseries-api/jobs/680120443#L498

similar to [what I'm seeing now](https://travis-ci.org/github/smartsdk/ngsi-timeseries-api/jobs/697246967#L498) in [build #1059](https://travis-ci.org/github/smartsdk/ngsi-timeseries-api/builds/697246963)---the latest. Am I missing something?

chicco785:
Nope, the build for "step-2" commit was working when the commit was made two months ago:

https://travis-ci.org/github/smartsdk/ngsi-timeseries-api/builds/680120428

to understand better i re-run and now it fails. this makes me think that the issue is related to the version update of some library

chicco785:
@c0c0n3 it could be alpine image, check in detail your build:
https://travis-ci.org/github/smartsdk/ngsi-timeseries-api/jobs/697246964#L230
```
Step 3/17 : RUN apk --no-cache --update-cache add gcc python python-dev py-pip build-base wget
 ---> Running in b558063cc519
fetch http://dl-cdn.alpinelinux.org/alpine/v3.12/main/x86_64/APKINDEX.tar.gz
fetch http://dl-cdn.alpinelinux.org/alpine/v3.12/community/x86_64/APKINDEX.tar.gz
ERROR: unsatisfiable constraints:
  python (missing):
    required by: world[python]
  python-dev (missing):
    required by: world[python-dev]
The command '/bin/sh -c apk --no-cache --update-cache add gcc python python-dev py-pip build-base wget' returned a non-zero code: 2
The command "source setup_dev_env.sh" exited with 0.
721.02s$ sh run_tests.sh
Using default tag: latest
latest: Pulling from smartsdk/quantumleap
c87736221ed0: Pulling fs layer
c3f51b0d0765: Pulling fs layer
b5a0acc6b737: Pulling fs layer
14428c09e432: Pulling fs layer
```

now check, one of the build working:

https://travis-ci.org/github/smartsdk/ngsi-timeseries-api/jobs/691676310#L227

change the alpine image to: 3.6.10-alpine3.11

and let's see

c0c0n3:
### The disease: dependency hell
> could be alpine image

Spot on, good catch! The problem change was this line in the Docker file:

    - FROM python:3.6.8-alpine3.8 as base
    + FROM python:3.6.10-alpine as base

which happened in **step 2**. How can it be then that the Travis build succeeded on step 2's commit (85d83bc)?! At the time of that commit `python:3.6.10-alpine` was pointing to a DockerHub image built from Alpine `3.10` as base whereas now the same tag points to an image built from Alpine `3.12`---the image maintainers [did the upgrade on 02 Jun 2020](https://github.com/docker-library/python/commit/ece154e2849e78c383419d0be591cfd332a471d3). The `apk` command in our Docker file doesn't work with Alpine `3.12` since Python packages have been renamed. This is why recent Travis builds were broken. On my machine `python:3.6.10-alpine` was pointing to a Docker image built from Alpine ` < 3.12` which is why I couldn't reproduce the issue---@chicco785 thanks for the tip!

After pinning the Alpine version and fixing up Alpine package names (6371199), I really thought I was out of the woods. No way Jose :-) All tests passed (on Travis too) but the Travis build still failed! In fact, `coveralls` bombed out on collecting coverage data after the test run. To be honest, I still haven't got the faintest why exactly that happened, but it looks like it could have something to do with conflicting Python libs. Here's what I could pick up. 85d83bc specified `python:3.6.10-alpine` so our QL image gets a Python `3.6.10` base environment (interpreter + base libs) in which `pipenv` installs all QL Python deps. Our `.travis.yml` tells Travis to run the build in a Python `3.6` base env and currently this is Python `3.6.7` on Travis. Also, it tells Travis to `pipenv install` before hand (`before_script` block), so the whole Travis build eventually runs in a Python `3.6.7` base env + QL libs that `pipenv` installs in there, in particular the `coveralls` client. So the coverage data gets collected in the Travis Python env which is slightly different than that of the QL image. I still can't see how that can possibly cause [this exception](https://travis-ci.org/github/smartsdk/ngsi-timeseries-api/jobs/698634012#L1433) though. Anyhoo, using the **exact same** Python version across the board (QL image, Travis, etc. see d315d5d) did the trick and now we have a clean build again, go figure!

### Home remedy
* Always prefer more specific Docker tags when available. In particular, we should always use `python:M.m.p-alpineN.n` rather than `python:M.m.p-alpine`---see 6371199. This doesn't really solve the problem since even e.g. `python:3.6.10-alpine3.12` will get updates from time to time but hopefully it'll always refer to the same Alpine version.
* Stop the Travis build in its tracks if Docker fails to build the QL image, so it should be easier to spot the problem going forward---see d4e6809.
* Use the exact same Python base env across the board: dev env, Travis, test suite, QL and Timescale setup images---see d315d5d. Ideally we should pin QL Python deps in the Pipfile too---not done yet.

### Bottom line
We've got a clean build again, but our Docker-based dependency management will get us from time to time into dependency hell. In fact, in general building and pulling of images are **time-dependent** operations. In other words the image you pull today from DockerHub for a given tag could be different than that you pulled yesterday for the same tag and likewise building the same Docker file twice might output different images. We should also make sure the various Python envs we juggle never get out of synch w/r/t each other.

### Notes
* **Alpine image**. We're on Alpine `3.12` now---see 6371199 and d315d5d.
* **Python**. Upgraded to latest stable version, `3.8.3`. Version `3.6` got the last security update in Dec 2019 (v `3.6.10`) and is no longer maintained AFAIK.
* **Dependency hell**. We've been bitten a few times before, e.g. #269, #270. The only way I know out of the mainstream messy approach to dependency management is to use Nix/NixOS or Guix/GuixSD. Thoughts?
* See also #273.

c0c0n3:
> Step 3 - understand if we got some free food:

No Sir, no free lunch :-) When using the Timescale backend, most tests fail. Some might fail cos of dirt left over from other tests that failed earlier---it's hard to tell and we'll need more time to find out. But I did try running some tests in isolation and they failed too...

> - explore changes needed to support timescale queries

I did some debugging to gauge what needs changing. It looks like for the most part, basic query building actually works fine---even with aggregates, yep `DATE_TRUNC` works!

Here are some of the queries QL ran on Timescale

    select entity_type,entity_id,time_index,"temperature" from "mtt1"."etroom"
    where  entity_id in ('Room0') and  fiware_servicepath = ''
    ORDER BY time_index ASC limit 10000 offset 0

    select entity_type,entity_id,count("temperature") as "temperature" from "mtt1"."etroom"
    where  entity_id in ('Room0') and  fiware_servicepath = ''
    GROUP BY entity_type,entity_id limit 10000 offset 0

    select entity_type,entity_id,time_index,"temperature" from "mtt1"."etroom"
    where  entity_id in ('Room0') and  
    time_index >= '1970-01-06 00:00:00+00:00'and  time_index <= '1970-01-17 00:00:00+00:00'
    and  fiware_servicepath = '' 
    ORDER BY time_index ASC limit 10000 offset 0

    select entity_type,entity_id,DATE_TRUNC('year',time_index) as time_index,
    avg("temperature") as "temperature"
    from "mtt1"."ettest_1t1e1a_aggrperiod"
    where  entity_id in ('test_1T1E1A_aggrPeriod0') and  fiware_servicepath = ''
    GROUP BY entity_type,entity_id,DATE_TRUNC('year', time_index)
    ORDER BY entity_type,entity_id,time_index ASC limit 10000 offset 0

They all went off without a hitch. What's really broken is the conversion of the Postgres result set into JSON. The code trips over when reading the metadata table, mainly because of implicit type conversions---e.g. the `col_names` passed to `_format_response` is a list of list of bytes whereas it looks like `_format_response` expects a list of strings so it fails to look up original attr info from the record fetched from the meta table.

I didn't have time to dig deeper, so there may be more issues. So I'm merging this PR and leave step 3 for another day :-)

chicco785:
> They all went off without a hitch. What's really broken is the conversion of the Postgres result set into JSON. The code trips over when reading the metadata table, mainly because of implicit type conversions---e.g. the `col_names` passed to `_format_response` is a list of list of bytes whereas it looks like `_format_response` expects a list of strings so it fails to look up original attr info from the record fetched from the meta table.
> 
> I didn't have time to dig deeper, so there may be more issues. So I'm merging this PR and leave step 3 for another day :-)

this is step-3 :) understand if we got free lunch or not!
I would say that if the problem is in the formatting the response, that's not so bad, I would be much more worried if the problem is in the query. Of course that's just an assumption, which may be totally wrong

PR #316, opened 2020-04-24
Cors

### LINKED ISSUE #135: How to manage CORS to consume data from QuantumLeap API
How disable CORS policy to consume data from QuantumLeap API.

cnoelle:
fixes #135

Adds CORS headers to responses. Configuration via env vars, e.g. QL_CORS_ALLOWED_HEADERS. See also https://fiware-orion.readthedocs.io/en/master/user/cors/index.html


Pipfile
```diff
@@ -8,6 +8,7 @@ certifi = "==2018.10.15"
 "connexion[swagger-ui]" = "~=2.2"
 crate = "~=0.22"
 flask = "~=1.1.0"
+flask_cors = "~=3.0.8"
```
c0c0n3, line 11: is flask cors `3.0.8` compatible with the rest of the web stack? i.e. flask `1.1.0`, connexion `2.2`, ...

src/app.py
```diff
@@ -1,12 +1,30 @@
 from utils.hosts import LOCAL
+from flask_cors import CORS
+import os
+
+
+def configure_cors(app, origin):
```
c0c0n3, line 6: excellent, thanks for putting this in a separate function. it'd be great if you could go the extra mile and use our [config reader](https://github.com/smartsdk/ngsi-timeseries-api/blob/master/src/utils/cfgreader.py) instead of plain python `os.env.get`. an example of how to use it is [over here](https://github.com/smartsdk/ngsi-timeseries-api/blob/77ed420c0a7532bcc13d941c0402f457cc40407a/src/translators/timescale.py#L81)

src/app.py
```diff
@@ -1,12 +1,30 @@
...
 if __name__ == '__main__':
     import connexion
+
     app = connexion.FlaskApp(__name__, specification_dir='../specification/')
     app.add_api('quantumleap.yml',
                 arguments={'title': 'QuantumLeap V2 API'},
                 pythonic_params=True,
                 # validate_responses=True, strict_validation=True
                 )
+    cors_allowed_origin = os.environ.get('QL_CORS_ALLOWED_ORIGIN', None)
```
c0c0n3, line 27: see my comment above about using our env reader util.

cnoelle:
Thanks for your comments, @c0c0n3 . I'll try to address them, in particular the tests may take a couple of days, though. When it is done I'll let you know.

chicco785:
no follow up in the last 6 months, this PR is closed.

PR #317, opened 2020-04-24
add a query for type ids

cnoelle:
This is a proposal for a new endpoint /v2/types, which simply returs the list of distinct entity type ids in the system. It seems that this functionality is not available yet.

There is also a default implementation available in the class BaseTranslator, which is probably not super efficient, but makes it work for all translators that implement query_ids. 

No tests yet, since I couldn't make pytest work on my machine :(

### DIFF
--- a/specification/quantumleap.yml
+++ b/specification/quantumleap.yml
@@ -1011,6 +1011,33 @@ paths:
                 "description": "No records were found for such query."
               }
 
+  /v2/types:
+    get:
+      operationId: reporter.reporter.get_types
+      summary: "Get all entity types present in the system"
+      description: "Get an array of distinct entity types present in the system."
+      tags:
+        - queries
+      parameters:
+        # In Query...
+        - $ref: '#/parameters/limit'
+        - $ref: '#/parameters/offset'
+        # In Header...
+        - $ref: '#/parameters/fiware-Service'
+        - $ref: '#/parameters/fiware-ServicePath'
+      responses:
+        200:
+          description: OK
+          schema:
+            type: object
+            properties:
+              types:
+                type: array
+                items:
+                  type: string
+          examples:
+            application/json:
+              { "types": ["Building", "Room", "TemperatureSensor"] }
 
   /v2/types/{entityType}/attrs/{attrName}:
     get:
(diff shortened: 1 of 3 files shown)


c0c0n3:
Hello @cnoelle! Thank you for the proposal.

I'm curious though, what's your use case? I'm struggling to find a good one. Is this endpoint supposed to help discover what kind of entities are in Quantum Leap? If so, how's that information going to be useful to the client? E.g. say

    GET /types

returned

    { "types": [ "X", "Y", "Z" ] }

How could the client use that information in any meaningful way without **prior** knowledge of what `X`, `Y` and `Z` are? Perhaps, the client knows all types in the DB have a common attribute? If so, how?

cnoelle:
Hi @c0c0n3 ,

the idea is exactly to discover the entity types present. I was thinking how a visualisation tool could allow the user to filter the data to be displayed. In a first step, it would display the available entity types, after selecting one of them the corresponding entity ids, then the attribute(s) of the selected id. The latter two filters can be implemented efficiently using the available REST endpoints (/v2/entities?type=<TYPE> and /v2/entities/<ENTITY_ID>), but the first one is missing.

cnoelle:
The client may also know certain entity types from a specified data model, e.g. https://fiware-datamodels.readthedocs.io/en/latest/Environment/AirQualityObserved/doc/spec/index.html.

c0c0n3:
> visualisation tool could allow the user to filter the data to be displayed

Ha! I wasn't thinking of the human sitting behind the screen. Yes, now it makes a lot of sense to me too, thanks for clarifying. Supposedly the person exploring your visualisation will have some knowledge of what those entity type names mean.

> types from a specified data model

I wouldn't bank on that, at least not until we support NGSI-LD. In practice you'll see data in all sorts of wonderful shapes :-)

One thing that springs to mind though is to use our metadata table to help the user pick a given entity type based on attributes of interest. Your query could also return the attribute names of those entities and as an added benefit, a plain query to the metadata table is going to be more efficient than calling `query_ids`. Killing two birds with one stone?

chicco785:
no follow up in the last 6 months, this PR is closed.

Issue #325, opened 2020-06-22
Log QL_DEFAULT_DB value when reading it from env


c0c0n3:
This is a left over of #315: I forgot to push this commit to the branch so it will have to go in a separate PR, sorry for the mess :-(

chicco785:
fix by #326

PR #326, opened 2020-06-22
Log env var values when reading it them in

c0c0n3:
Implements #325.

### DIFF
--- a/src/geocoding/factory.py
+++ b/src/geocoding/factory.py
@@ -1,0 +1,70 @@
+import logging
+from typing import Union
+
+from .geocache import GeoCodingCache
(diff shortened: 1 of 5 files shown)


c0c0n3:
So I've implemented both #325 and #327. In the process I've also factored out the geocoding setup logic that was scattered about the `reporter` and `health` modules with slightly different implementations:

* `reporter` would only use geocoding if the use geocoding var was true and the redis host var was set;
* `health` would check redis and osm if the use geocoding var was true.

So if `use geocoding = true` but `redis not set`, the health checks would run for nothing since the `reporter` wouldn't use geocoding anyway. This is fixed now and the setup logic sits in just one place---see new `geocoding.factory` module.

chicco785:
LGTM

PR #333, opened 2020-07-04
(ja) Update Japanese documents

fisuda:
This PR updates some Japanese documents in docs/manuals.ja. It would be great if you could review this PR.


c0c0n3:
@fisuda that's great, thanks for this!! I'm going to merge this PR just now...

Issue #334, opened 2020-07-06
QL doesn't notify when CrateDB is locked due to less available memory


ccsr:
When the allocated available memory to CrateDB goes below 15%, CrateDB gets locked and QL doesn't give any errors.

1. Create a subscription from Context Borker to QL and wait till the allocated free disk is below 15%
2. CrateDB locks and QL don't give any errors

**Expected behavior**
QL should return "not processed or not updated" error code to subscription call

chicco785:
@ccsr thx, we will check

daminichopra:
Hi @chicco785 @ccsr @c0c0n3 , I investigated on this issue and below are the error logs which i get while notifying request.

![Capture](https://user-images.githubusercontent.com/35693908/88383492-400d7b00-cdc8-11ea-9019-de9b70194d91.PNG)
![capture1](https://user-images.githubusercontent.com/35693908/88383501-44d22f00-cdc8-11ea-8f88-e66eb44cc0b0.PNG)

Please confirm and assign this issue to me if nobody is working on this Issue. Thank you :)

chicco785:
@daminichopra assigned!

c0c0n3:
closed by #346

PR #346, opened 2020-08-04
Updated QL to notify when memory not available

daminichopra:
To update message from quantum leap when database memory is not available with "not processed or not updated"

### DIFF
--- a/src/reporter/reporter.py
+++ b/src/reporter/reporter.py
@@ -191,7 +191,7 @@ def notify():
     except:
         msg = "Notification not processed or not updated"
         log().info(msg)
-        return msg,500
+        return msg, 500
     msg = 'Notification successfully processed'
     log().info(msg)
     return msg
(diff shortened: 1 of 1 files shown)


daminichopra:
Fix for  [Issue](https://github.com/smartsdk/ngsi-timeseries-api/issues/334).

src/reporter/reporter.py
```diff
@@ -185,9 +185,13 @@ def notify():
...
     # Send valid entities to translator
-    with translator_for(fiware_s) as trans:
-        trans.insert(payload, fiware_s, fiware_sp)
-
+    try:
+        with translator_for(fiware_s) as trans:
+            trans.insert(payload, fiware_s, fiware_sp)
+    except:
+        msg = "Notification not processed or not updated"
+        log().info(msg)
```
c0c0n3, line 193: can we change the log level from info to error? thanks!! :-)

c0c0n3:
@daminichopra thanks a stack for wrapping up this PR, I'm going to merge it :-)

PR #354, opened 2020-08-25
WIP: load testing and gunicorn support

chicco785:


c0c0n3:
Just for future reference, the "cleaned up" version of this PR sits in the following commits of the `load-test` and `gunicorn` branches

* *port notify load test from #354.* (a13a8f5) This commit basically lifts the load test as is from this PR and replays it on master.
* *make sure load test uses latest ql code and removes containers to avoid undeletable dangling images.* (960a148) This is a small fix to the load test to avoid problems w/ dangling docker images.
* *port gunicorn setup from #354.* (6b84e35) This commit basically lifts the Gunicorn setup as is from this PR and replays it on master.
* *get rid of supervisord.* (4c6f1b1) This is a small improvement over the previous commit since supervisord isn't actually needed. In fact, it'd only monitor the Gunicorn master process which already takes care of monitoring child worker processes. Also Quantum Leap logs were not output to stdout. (This is a consequence of our supervisord config.)

As agreed, we left out the caching stuff, i.e. following changes in this PR

* [src/tests/common.py](https://github.com/smartsdk/ngsi-timeseries-api/pull/354/files#diff-e95a3ca1cecfb89f780a720add3555b4)
* [src/translators/crate.py](https://github.com/smartsdk/ngsi-timeseries-api/pull/354/files#diff-038dc5c6921c5e01d1f89c6db1778a13)
* [src/translators/sql_translator.py](https://github.com/smartsdk/ngsi-timeseries-api/pull/354/files#diff-29ca943e9e15c068d720d9007a0bfddd)
* [src/translators/table_cache.py](https://github.com/smartsdk/ngsi-timeseries-api/pull/354/files#diff-6779dbeebac5dfe40c87276964b85675)
* `src/translators/tests/*.py`, i.e. all pytests in the translator package.

Issue #362, opened 2020-09-14
Gunicorn container halts test session


c0c0n3:
### Describe the bug
During the test session in the Travis build, it looks like the Gunicorn server in the Docker container suddenly comes to the foreground, so tests stop, and eventually Travis kills the build.

### To Reproduce
Do a clean check out of the `master` branch, **fully** prune your Docker system taking care of not leaving behind any dangling images/containers, remove all your Python venvs, clean up any Python env vars, then in a fresh terminal:

    $ cd where/you/cloned/ngsi-timeseries-api
    $ pipenv install
    $ source setup_dev_env.sh
    $ sh run_tests.sh 

At some point during the test session you should see the Gunicorn container coming to the foreground and making the tests hang.

### Expected behavior
Gunicorn container should sit in the background, allowing `pytest` to go on with the test session.

### Additional context
See any of the following build logs:

- https://travis-ci.org/github/smartsdk/ngsi-timeseries-api/builds/727051434
- https://travis-ci.org/github/smartsdk/ngsi-timeseries-api/builds/726970500

chicco785:
the problem is not gunicorn, but the new image using entrypoint instead of command, which makes the integration tests fail in the content load fase

c0c0n3:
ah, thanks for that. I wonder why tests were passing for a while even on Travis as well as on our boxes. Anyhoo, let's fix forward.

c0c0n3:
fixed by #363

Issue #347, opened 2020-08-04
Programmatic deletion of Table from Crate DB 


ilarimikkonen:
**Is your feature request related to a problem? Please describe.**
When data is received by QL via notifications from CB, a table is created to Crate DB. Data can be deleted from Crate via API endpoints, but the table stays in the Crate DB. 

**Describe the solution you'd like**
An API endpoint to remove table from the Crate DB

**Describe alternatives you've considered**
1) exec to container 2) start Crash 3) DROP TABLE
This is not convinient for some use cases, API endpoint would be good for programmatic access.

**Additional context**
Over time, tables can accumulate to the Crate DB and this new API endpoint would help with house keeping.

c0c0n3:
> Data can be deleted from Crate via API endpoints, but the table stays in the Crate DB.

If you delete **all** records through the API delete endpoint, then this should happen automatically:

1. The entity table gets dropped---i.e. QL should issue the `DROP TABLE ...` statement for you.
2. The corresponding table entry should be deleted from the metadata table.

It looks like this isn't happening for you? Can you confirm you deleted **all** the records in the table **using the delete API**? Thanks!

ilarimikkonen:
Setup:
Running Crate 3.3.5. with 0.7.5. QL. Orion in front

Test:
1) create a new tenant: deletetest1 - OK
2) push data to tenant - OK
3) get data from tenant - OK
4) create a subscription to QL - OK
5) push data - OK
6) check subscription - OK
7) check data from Crate - OK:
```
cr> select * from  "mtdeletetest1"."etdeletetest";
+------------+-------------+--------------------+---------------+-------+
| entity_id  | entity_type | fiware_servicepath |    time_index | value |
+------------+-------------+--------------------+---------------+-------+
| deletetest | deletetest  | /                  | 1597135006000 |   1.0 |
| deletetest | deletetest  | /                  | 1597135138000 |   2.0 |
+------------+-------------+--------------------+---------------+-------+
SELECT 2 rows in set (0.006 sec)
cr>
```
8) do delete - OK:
```
curl --location --request DELETE 'https://sthdata.testurlhere/ql/v2/types/deletetest' \
--header 'fiware-servicepath: /' \
--header 'fiware-service: deletetest1' 
```
9) check from crate - NOK:
```
cr> select * from  "mtdeletetest1"."etdeletetest";
+-----------+-------------+--------------------+------------+-------+
[...]

PR #363, opened 2020-09-19
fix #208

### LINKED ISSUE #208: issues with multi entity notification (wrong servicepath created)
**Describe the bug**
When a multiple notification is sent, the service path created is a comma separated version of all servicepaths of the entities.
It could be that the header sent by orion is exactly that. We should then split it and attach it one for entity.

entity_id | entity_type | fiware_servicepath | powerstate | time_index | timeinstant
-- | -- | -- | -- | -- | --
urn:ngsi-ld:Streetlight:sl-6 | Streetlight | /Diorama,/Diorama,/Diorama,/Diorama,/Diorama,/Diorama | NULL | 1560425955297 (2019-06-13T11:39:15.297Z) | NULL
urn:ngsi-ld:Streetlight:sl-1 | Streetlight | /Diorama,/Diorama,/Diorama,/Diorama,/Diorama,/Diorama | NULL | 1560422941000 (2019-06-13T10:49:01.000Z) | 1560422941000 (2019-06-13T10:49:01.000Z)
urn:ngsi-ld:Streetlight:sl-3 | Streetlight | /Diorama,/Diorama,/Diorama,/Diorama,/Diorama,/Diorama | NULL | 1560422941000 (2019-06-13T10:49:01.000Z) | 1560422941000 (2019-06-13T10:49:01.000Z)
urn:ngsi-ld:Streetlight:sl-2 | Streetlight | /Diorama,/Diorama,/Diorama,/Diorama,/Diorama,/Diorama | NULL | 1560422941000 (2019-06-13T10:49:01.000Z) | 1560422941000 (2019-06-13T10:49:01.000Z)
urn:ngsi-ld:Streetlight:sl-5 | Streetlight | /Diorama,/Diorama,/Diorama,/Diorama,/Diorama,/Diorama | NULL | 1560425955297 (2019-06-13T11:39:15.297Z) | NULL
urn:ngsi-ld:Streetlight:sl-4 | Streetlight | /Diorama,/Diorama,/Diorama,/Diorama,/Diorama,/Diorama | NULL | 1560360355000 (2019-06-12T17:25:55.000Z) | 1560360355000 (2019-06-12T17:25:55.000Z)

**Environment (please complete the following information):**
 - CrateDB 3.3.3

chicco785:
* added code to group entities by servicepath when multiple notifications arrive at once (#208)
* add tests and integration tests (so we have also test with orion triggering such case)
* fix notification tests to rely only on api calls (remove call to translator)
* changed error code to 400 when due to invalid request
* upgraded python to 3.8.5
* fix integration test to use new docker image with entrypoint (#362)

### DIFF
--- a/src/tests/run_load_tests.sh
+++ b/src/tests/run_load_tests.sh
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+docker build --cache-from smartsdk/quantumleap -t smartsdk/quantumleap ../../
+
 docker-compose up -d
 docker-compose stop orion
 docker-compose stop mongo
--- a/src/tests/run_load_tests.sh
+++ b/src/tests/run_load_tests.sh
@@ -1,7 +1,5 @@
 #!/usr/bin/env bash
 
-docker build --cache-from smartsdk/quantumleap -t smartsdk/quantumleap ../../
-
 docker-compose up -d
 docker-compose stop orion
 docker-compose stop mongo
@@ -15,4 +13,4 @@ docker run -i --rm loadimpact/k6 run --vus 100 --duration 120s - < notify-load-t
 
 sleep 10
 
-docker-compose down
+docker-compose down -v
--- a/src/translators/sql_translator.py
+++ b/src/translators/sql_translator.py
@@ -1,7 +1,7 @@
 from datetime import datetime, timedelta, timezone
 from geocoding.slf.geotypes import *
 from exceptions.exceptions import AmbiguousNGSIIdError, UnsupportedOption, \
-    NGSIUsageError, InvalidParameterValue
+    NGSIUsageError, InvalidParameterValue, InvalidHeaderValue
 from translators import base_translator
 from translators.config import SQLTranslatorConfig
 from utils.common import iter_entity_attrs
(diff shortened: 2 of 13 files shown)


src/translators/sql_translator.py
```diff
@@ -170,16 +170,46 @@ def insert(self, entities, fiware_service=None, fiware_servicepath=None):
...
+                    res = self._insert_entities_of_type(et,
+                                                        entities_by_type[et],
+                                                        fiware_service,
+                                                        path)
+        else:
+            msg = 'Multiple servicePath are allowed only ' \
+                  'if their size is match the size of entities'
+            raise InvalidHeaderValue('Fiware-ServicePath',
+                                     fiware_servicepath, msg)
+
```
c0c0n3, line 212: what happens if there's only one entity and an empty service path?

    service_paths = ['']
    for idx, path in enumerate(service_paths):
        print(f"{idx}: {path}")

    >>>     0:

so I think we'd be inserting the entity with an empty path? In general, do we still allow clients not to specify a service path, so e.g. a notification w/ 3 entities and no service path results in all three entities still being inserted under the empty path?

src/tests/run_load_tests.sh
```diff
@@ -1,7 +1,5 @@
 #!/usr/bin/env bash
 
-docker build --cache-from smartsdk/quantumleap -t smartsdk/quantumleap ../../
```
c0c0n3, line 3: don't we need this line? I think w/o it, when you run the script you get the latest image in the docker registry which doesn't not necessarily reflects the latest code changes? for example, say I check out a branch someone is working on, then run this script. The tests will run with my local image, which won't have the changes in the branch?

src/translators/sql_translator.py
chicco785: may the correct one is `service_paths = [None]` 
all actually boils down to what orion provide us, right?

src/tests/run_load_tests.sh
chicco785: correct, but for load test i thought it was not needed. can be inserted back

src/translators/sql_translator.py
chicco785: see line 178, so if no service path, we exactly create an array with `service_paths = [None]`

c0c0n3: yep, line 178 is what makes a list with one entry if there's no service path (`None` or `''`). just wanted to double check this was intentional.

> boils down to what orion provide us, right?

yep, you're right. I don't think we've got much of a choice here. If the notification contains no service path probably we should still insert the entity. Looking at the code I think we do that in the case of one entity, but what does Orion do when notifying multiple entities with no service path? Will the notification contain a list of empty service paths, e.g. for 3 entities

    path = ' , , '
    entities = [e1, e2, e3]

or no path at all, e.g.

    path = ''
    entities = [e1, e2, e3]

If I read the code correctly, in the first case we insert all 3 entities under the empty path whereas in the second case we raise an `InvalidHeaderValue` exception. If this is the case, then my question is: do we have to insert entities if there's no service path or it is empty?

src/tests/run_load_tests.sh
c0c0n3: cool, we could put it back in after rebasing #347 on master? or is it easier to do in this PR?

src/translators/sql_translator.py
chicco785: Multiple "" will work if they are in the same number as entities.

c0c0n3 (approved):

PR #370, opened 2020-09-24
Improve crate 4.x support 

chicco785:
Since cratedb 4.0 timestamp is deprecated and may be removed any time soon. This update, provide timestamptz support for cratedb >= 4.0

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ QuantumLeap supports both Crate DB and Timescale as time-series DB
 backends but please bear in mind that at the moment we only support
 the following versions:
 
-* Crate backend: Crate DB version `3.3.*` and `4.*` (experimental)
+* Crate backend: Crate DB version `3.3.*` (will be deprecated from QL `0.8` version) and `4.*`
 * Timescale backend: Postgres version `10.*` or `11.*` +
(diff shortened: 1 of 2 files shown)


c0c0n3:
Same applies to other data types we use:

* https://crate.io/docs/crate/reference/en/4.2/appendices/release-notes/4.0.0.html#general

Deprecated types have been aliased so nothing is going to break for the next couple of releases I'd imagine, but since we're at it, I suggest we take care of the other types too---`string`, `long` and not sure what else?

chicco785:
> Same applies to other data types we use:
> 
> * https://crate.io/docs/crate/reference/en/4.2/appendices/release-notes/4.0.0.html#general
> 
> Deprecated types have been aliased so nothing is going to break for the next couple of releases I'd imagine, but since we're at it, I suggest we take care of the other types too---`string`, `long` and not sure what else?

not a problem, but there is a substantial difference, timestamp generates billions of logs in crate, the others don't.

c0c0n3:
> not a problem

I saw the new commit, good stuff!

> timestamp generates billions of logs in crate, the others don't.

Shaw! I wonder why.

c0c0n3 (approved):

PR #371, opened 2020-09-29
fix date format to be compatible also with crate 3.x (#353)

chicco785:
#353 reported `fromDate` and `toDate` to not be working correctly. test we run using crate 4.x didn't expose the problem. still user reported the issue, so we looked into using crate 3.x (that will be deprecated in 0.8!) and have been able to reproduce the issue. while the generated timestamp (without `T`) was supported by crate 4.x, it is not the case for 3.x. we changed the timestamp to include `T`.

the issue didn't surface because of the related exception is `managed`, this should probably be modified. currently, the error is at least promoted from "debug" to "error" log, but probably the exception should be simply raised and correctly reported to api caller.

on an additional note, it is unclear why the test using old version of crate that we run on each PR didn't break.

### DIFF
--- a/src/translators/sql_translator.py
+++ b/src/translators/sql_translator.py
@@ -520,6 +520,7 @@ class SQLTranslator(base_translator.BaseTranslator):
     def _get_where_clause(self, entity_ids, from_date, to_date, fiware_sp=None,
                           geo_query=None):
         clauses = []
+        where_clause = ""
 
         if entity_ids:
             ids = ",".join("'{}'".format(e) for e in entity_ids)
@@ -533,8 +534,12 @@ class SQLTranslator(base_translator.BaseTranslator):
 
         if fiware_sp:
             # Match prefix of fiware service path
-            clauses.append(
-                " " + FIWARE_SERVICEPATH + " ~* '" + fiware_sp + "($|/.*)'")
+            if fiware_sp == '/':
+                clauses.append(
+                    " " + FIWARE_SERVICEPATH + " ~* '($|/.*)'")
+            else:
+                clauses.append(
+                    " " + FIWARE_SERVICEPATH + " ~* '" + fiware_sp + "($|/.*)'")
         else:
             # Match prefix of fiware service path
             clauses.append(" " + FIWARE_SERVICEPATH + " = ''")
@@ -543,12 +548,13 @@ class SQLTranslator(base_translator.BaseTranslator):
         if geo_clause:
             clauses.append(geo_clause)
 
-        where_clause = "where " + "and ".join(clauses)
+        if len(clauses) > 0:
+            where_clause = "where" + " and ".join(clauses)
         return where_clause
 
     def _parse_date(self, date):
         try:
-            return dateutil.parser.isoparse(date.strip('\"'))
+            return dateutil.parser.isoparse(date.strip('\"')).isoformat()
         except Exception as e:
             raise InvalidParameterValue(date, "**fromDate** or **toDate**")
 
@@ -805,8 +811,10 @@ class SQLTranslator(base_translator.BaseTranslator):
             try:
                 self.cursor.execute(op)
             except Exception as e:
+                #TODO due to this except in case of sql errors,
+                #all goes fine, and users gets 404 as result
                 # Reason 1: fiware_service_path column in legacy dbs.
-                logging.debug("{}".format(e))
+                logging.error("{}".format(e))
                 entities = []
             else:
                 res = self.cursor.fetchall()
--- a/src/translators/sql_translator.py
+++ b/src/translators/sql_translator.py
@@ -536,7 +536,7 @@ class SQLTranslator(base_translator.BaseTranslator):
             # Match prefix of fiware service path
             if fiware_sp == '/':
                 clauses.append(
-                    " " + FIWARE_SERVICEPATH + " ~* '($|/.*)'")
+                    " " + FIWARE_SERVICEPATH + " ~* '/.*'")
             else:
                 clauses.append(
                     " " + FIWARE_SERVICEPATH + " ~* '"
--- a/src/translators/sql_translator.py
+++ b/src/translators/sql_translator.py
@@ -539,7 +539,8 @@ class SQLTranslator(base_translator.BaseTranslator):
                     " " + FIWARE_SERVICEPATH + " ~* '($|/.*)'")
             else:
                 clauses.append(
-                    " " + FIWARE_SERVICEPATH + " ~* '" + fiware_sp + "($|/.*)'")
+                    " " + FIWARE_SERVICEPATH + " ~* '"
+                    + fiware_sp + "($|/.*)'")
         else:
             # Match prefix of fiware service path
             clauses.append(" " + FIWARE_SERVICEPATH + " = ''")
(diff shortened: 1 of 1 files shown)


src/translators/sql_translator.py
```diff
@@ -533,8 +534,13 @@ def _get_where_clause(self, entity_ids, from_date, to_date, fiware_sp=None,
 
         if fiware_sp:
             # Match prefix of fiware service path
-            clauses.append(
-                " " + FIWARE_SERVICEPATH + " ~* '" + fiware_sp + "($|/.*)'")
+            if fiware_sp == '/':
+                clauses.append(
+                    " " + FIWARE_SERVICEPATH + " ~* '($|/.*)'")
```
chicco785, line 539: previous query was not working correctly with "/" subpaths

src/translators/sql_translator.py
```diff
@@ -543,12 +549,13 @@ def _get_where_clause(self, entity_ids, from_date, to_date, fiware_sp=None,
         if geo_clause:
             clauses.append(geo_clause)
 
-        where_clause = "where " + "and ".join(clauses)
+        if len(clauses) > 0:
+            where_clause = "where" + " and ".join(clauses)
```
chicco785, line 553: from code logic clauses will be always > 0, but if anything changes that, this would ensure that query does not break

src/translators/sql_translator.py
```diff
@@ -543,12 +549,13 @@ def _get_where_clause(self, entity_ids, from_date, to_date, fiware_sp=None,
...
         return where_clause
 
     def _parse_date(self, date):
         try:
-            return dateutil.parser.isoparse(date.strip('\"'))
+            return dateutil.parser.isoparse(date.strip('\"')).isoformat()
```
chicco785, line 558: in this way the we get the correct iso format compatible with both crate 3.x and crate 4.x

src/translators/sql_translator.py
```diff
@@ -805,8 +812,10 @@ def query(self,
             try:
                 self.cursor.execute(op)
             except Exception as e:
+                # TODO due to this except in case of sql errors,
+                # all goes fine, and users gets 404 as result
                 # Reason 1: fiware_service_path column in legacy dbs.
-                logging.debug("{}".format(e))
+                logging.error("{}".format(e))
```
chicco785, line 818: for now we just promoted the log to error, but i think it would be correct just to raise the exception to the api call level

src/translators/sql_translator.py
c0c0n3: Not sure about this change. Let's start with an example.

```
create table t (path text);
insert into t values (''), ('a'), ('a/b'), ('a/b/c'), ('/'), ('/a'), ('/a/b'), ('/a/b/c');
select * from t where path ~* '($|/.*)';
```

this selects five paths: the empty path `''` and the four "absolute" (i.e. starting with a `/`) paths `/`, `/a`, `/a/b`, `/a/b/c`. Judging by the code, the empty path is sort of different from any other path in that it should only match an empty path in the DB? If so then an input path of `/` should not match the empty path...or should it? It looks to me the intention here is to match an input path `x` against the set of DB paths `FSP` as below

```
(1)   x == None or x == ''         ===>   { '' }
(2)   len(x) > 0 and x[0] != '/'   ===>   { p ∈ FSP | p starts with: '/' + x }
(3)   len(x) > 0 and x[0] == '/'   ===>   { p ∈ FSP | p starts with: x } 
```

In words, if there's no input path (1), pick up any records having an empty path. Otherwise (2) if the input path is relative (doesn't start with `/`), make it absolute (prepend `/`) and select any records whose path starts with it. Finally (3) if the path is absolute, collect any records whose path starts with it.

Sure, I can be wrong. I'm just trying to deduce a spec (which we never had!) from the code. IMHO, we should first clarify what the heck a service path is. It looks to me it's just an encoding of a path on a tree? Perhaps a tree used to organise devices in a hierarchy? A formal definition would help here. Once we have that, we can give a precise meaning to "matching a service path"...

src/translators/sql_translator.py
c0c0n3: yep, much better!

src/translators/sql_translator.py
c0c0n3: great, I'll have to see what happens with Timescale. I'll wait until we merge this PR.

src/translators/sql_translator.py
c0c0n3: yea, we should. Also, I've got no clue what that comment (reason 1) actually means. Does it ring a bell to you?

src/translators/sql_translator.py
chicco785: A service path is a hierarchy: https://fiware-orion.readthedocs.io/en/master/user/service_path/index.html

In the previous implementation, `/a` would select all child of `/a` path, i.e. `/a/b` and `/a/b/c`. The previous query does not work correctly (and return no result) if you ask for `/`, which assuming we want to be consistent, is incorrect.

I would be in favour of introducing `/#` as in context broker, but that would be quite deep change and need to be announced, since people may now rely on `/a` to return subpaths. So goal now it is just to be consistent. if `/a` returns `/a` and `/a/b`, `/` should return `/a`, `/b` and `/a/b`

src/translators/sql_translator.py
chicco785: i think it was something introduced by tomas, I don't have clue about it. a simple approach could be raise the exception, and see if tests work fine or breaks. if work fine, we go that way now, if not we keep the log, and later fix to introduce proper error handling

src/translators/sql_translator.py
chicco785: '(/.*)' is the correct filter for the second case

c0c0n3: yep, it looks right to me too, but I haven't tested it

chicco785: > > https://fiware-orion.readthedocs.io/en/master/user/service_path/index.html
> 
> Ouch. Not the kind of spec I was hoping for. After reading it I've got even more questions than before. There's no clear definition of those terms and it actually looks like that to them, depending on the context, a hierarchy could be a tree or a directed graph (an entity could be under two paths?) but was probably meant to be a poset. Me head hurts :-) Let's forget I ever asked the question and try to fix this based on what QL used to do :-)

there may be two entities with the same id under two different paths, but this are logically two different entities. i think it's explained somewhere. so still a tree, not a graph.
 
> Okay, so `/` shouldn't match the empty path? Or am I missing something?

it should not (also, it should not be possible to have empty path, according to context broker specs, minimal path is always `/`)

c0c0n3: > here may be two entities with the same id under two different paths, but this are logically two different entities. i think it's explained somewhere. so still a tree, not a graph.

Well, that's one **interpretation** :-) As a matter of fact there's no precise definition and quite a number of statements are seemingly contradictory, e.g. *"It is possible to have an entity with the same ID and type in different Scopes"*, *"Entities belongs to one (and only one) scope."*. Have you seen any statements along the lines of e.g. "a hierarchy is a tree whose leaf nodes are devices...",  "a service path is a string encoding of a root to leaf path whereby..." I don't think we can rely on that spec. To me we should just treat the service path as a sort of entity label which we handle exactly how QL has been doing so far + your root path fix?

> it should not (also, it should not be possible to have empty path, according to context broker specs, minimal path is always /)

What should we do here? We could reject empty service paths when inserting entities but that would be one hell of a breaking change...

src/translators/sql_translator.py
chicco785: tested, things break because it seems that after you drop tables (so they are not there), having still content in the metadata table (probably due to needed refresh time), ql attempts to query them and bombs.

c0c0n3: cool, thanks for this!! let's keep the log for now then. I've opened an issue about it just in case we want to revisit this scenario in the future: #372.

c0c0n3 (approved):
one push build failed b/c of an issue unrelated to this pr---apk failure. but all pr build were okay, ready to merge! 👍

PR #373, opened 2020-10-01
redis cache on insert

chicco785:
quantumleap heavily relies on crate (or timescale) to query and store data model configuration, this implies during insert an high number of queries on the backend, despite the change of models between an insert and another of the same entity type should not be that frequent.

as tested, caching reduce such number of queries, and thus increase throughput. previous test used in-memory cache, but that's not ideal in a concurrent environment, thus in this new test we experiment the usage of redis.

experiments shows that:
* caching metadata improves throughput of inserts , but usage of redis, has not the same impact as in memory
* the number of queries linked to metadata is not the only element affecting insert performance, due to the nature of http request so far we create a db connection for each insert, opening the connection takes time. in this pr we cache the connection, this proved to increase throughput of 100% compared just to just redis caching.

caveats: if you use caching, manual delete (off course) can make query bombs unless you clear also cache or wait cache to be expired

System used for testing:
Mac2016 with 3.1 GhZ i7 (i.e. 4 cores) and 16GB Ram

* Baseline (latest image as off 2020/10/02): 44 req/s - 600 ms avg response time -  (crate queries peak: select 111 q/sec, insert 54 q/sec)
* Redis caching: 60 req/s -  (no data on response time - have to check old image)
* In memory caching:  95 req/s - (no data on response time - have to check old image)
* Redis caching + connection re-usage: 100 req/s - 200 ms avg response time (crate queries peak: select 7 q/sec, insert 142 q/sec)
* Connection re-usage (without any cache): 55 req/s  - 400 ms avg response time (crate queries peak: select 177 q/sec, insert 86 q/sec)

* Redis caching + connection re-usage data:

```
docker run -i --rm loadimpact/k6 run --vus 30 --duration 60s - < notify-load-test.js

    checks.....................: 100.00% ✓ 6000 ✗ 0   
    data_received..............: 1.6 MB  27 kB/s
    data_sent..................: 1.6 MB  27 kB/s
    http_req_blocked...........: avg=302.26µs min=1.62µs      med=6.21µs   max=1.05s   p(90)=10.21µs  p(95)=18µs    
    http_req_connecting........: avg=262.78µs min=0s          med=0s       max=1.05s   p(90)=0s       p(95)=0s      
    http_req_duration..........: avg=208.12ms min=16.92ms     med=155.56ms max=1.06s   p(90)=409.38ms p(95)=514.53ms
    http_req_receiving.........: avg=1.2ms    min=-8.817564ms med=158.96µs max=133.5ms p(90)=2.69ms   p(95)=5.99ms  
    http_req_sending...........: avg=89.36µs  min=12.59µs     med=38.08µs  max=20.85ms p(90)=111.97µs p(95)=218.72µs
    http_req_tls_handshaking...: avg=0s       min=0s          med=0s       max=0s      p(90)=0s       p(95)=0s      
    http_req_waiting...........: avg=206.82ms min=16.78ms     med=153.98ms max=1.06s   p(90)=406.54ms p(95)=513.29ms
    http_reqs..................: 6000    99.814509/s
    iteration_duration.........: avg=30.02s   min=30.02s      med=30.02s   max=30.03s  p(90)=30.02s   p(95)=30.02s  
    iterations.................: 60      0.998145/s
    vus........................: 30      min=30 max=30
    vus_max....................: 30      min=30 max=30
```


chicco785:
the env variable to set ttl for cache has to be documented. default is 60 sec.

src/translators/crate.py
```diff
@@ -203,3 +206,28 @@ def CrateTranslatorInstance():
...
+
+    def __init__(self):
+        if not Borg.__monostate:
+            Borg.__monostate = self.__dict__
+            # Your definitions here
+            self.connection = None
+        else:
+            self.__dict__ = Borg.__monostate
+
+class CrateConnectionManager(Borg):
```
chicco785, line 222: so far only for crate

src/translators/sql_translator.py
```diff
@@ -946,8 +969,20 @@ def _format_response(self, resultset, col_names, table_name, last_n):
...
+        try:
+            # TODO we tested using cache here, but with current "delete" approach this causes issues
+            # scenario triggering the issue is: a entity is create, delete is used to delete all values
+            # what happens is that the table is empty, but metadata are still there, so caching the query with
+            # res = self._execute_query_via_cache(table_name, "metadata", stmt, [table_name], self.default_ttl)
+            # actually create an entry in the cache table_name, "metadata"
+            # in a following query call (below ttl) the same cache can be called despite there is no data.
+            # a possible solution is to create a cache based on query parameters that would cache all the results
+            self.cursor.execute(stmt, [table_name])
+            res = self.cursor.fetchall()
```
chicco785, line 981: we tested using cache here, but with current "delete" approach this causes issues scenario triggering the issue is: a entity is create, delete is used to delete all values what happens is that the table is empty, but metadata are still there, so caching the query with
```
res = self._execute_query_via_cache(table_name, "metadata", stmt, [table_name], self.default_ttl)
```
actually create an entry in the cache table_name, "metadata", in a following query call (below ttl) the same cache can be called despite there is no data. a possible solution is to create a cache based on query parameters that would cache all the results

src/translators/sql_translator.py
```diff
@@ -1060,6 +1105,50 @@ def drop_table(self, etype, fiware_service=None):
             except Exception as e:
                 logging.error("{}".format(e))
 
+    def query_entity_types(self, fiware_service=None, fiware_servicepath=None):
+        """
```
chicco785, line 1109: baseline to be extended to provide /v2/types endpoint ;)

src/translators/sql_translator.py
```diff
@@ -1105,3 +1197,68 @@ def _get_entity_type(self, entity_id, fiware_service):
 
     def _compute_type(self, attr_t, attr):
         raise NotImplementedError
+
+    def _execute_query_via_cache(self, table_name, key, stmt, parameters=None,
+                                 ex=None):
```
chicco785, line 1202: this function provide a way to check if a query is in cache in a certain key, table tuple. if not execute the query.

src/translators/sql_translator.py
```diff
@@ -1105,3 +1197,68 @@ def _get_entity_type(self, entity_id, fiware_service):
...
+                logging.warning(e.__str__())
+
+        self.cursor.execute(stmt, parameters)
+        res = self.cursor.fetchall()
+        if res:
+            self._cache(table_name, key, res, ex)
+        return res
+
+    def _is_query_in_cache(self, table_name, key):
+        if self.cache:
```
chicco785, line 1217: check if a key, table tuple is in cache

src/translators/sql_translator.py
```diff
@@ -1105,3 +1197,68 @@ def _get_entity_type(self, entity_id, fiware_service):
...
+            try:
+                return self.cache.exists(table_name, key)
+            except Exception as e:
+                logging.warning(e.__str__())
+        return False
+
+    def _cache(self, table_name, key, value=None, ex=None):
+        if self.cache:
```
chicco785, line 1225: cache a certain value in a tuple table, key for tot seconds

src/translators/sql_translator.py
```diff
@@ -1105,3 +1197,68 @@ def _get_entity_type(self, entity_id, fiware_service):
...
+            try:
+                if value:
+                    value = pickle.dumps(value)
+                self.cache.put(table_name, key, value, ex)
+            except Exception as e:
+                logging.warning(e.__str__())
+
+    def _remove_from_cache(self, table_name, key):
+        if self.cache:
```
chicco785, line 1234: remove a tuple table, key from cache

src/translators/sql_translator.py
```diff
@@ -1105,3 +1197,68 @@ def _get_entity_type(self, entity_id, fiware_service):
...
+        if not Borg.__monostate:
+            Borg.__monostate = self.__dict__
+            # Your definitions here
+            self.cache = None
+        else:
+            self.__dict__ = Borg.__monostate
+
+
+class QueryCacheManager(Borg):
+
```
chicco785, line 1254: singleton to not recreate connection to cache at every request

src/translators/crate.py
chicco785, line 222: singleton to not recreate connection to db at every request

src/translators/sql_translator.py
```diff
@@ -305,9 +316,10 @@ def _insert_entities_of_type(self,
 
         # Create/Update metadata table for this type
         table_name = self._et2tn(entity_type, fiware_service)
-        self._update_metadata_table(table_name, original_attrs)
+        modified = self._update_metadata_table(table_name, original_attrs)
         # Sort out data table.
-        self._prepare_data_table(table_name, table, fiware_service)
+        if modified:
+            self._prepare_data_table(table_name, table, fiware_service)
```
chicco785, line 322: now `_update_metadata_table` return 0 if no difference between existing metadata and new ones, if difference it return 1, this allows to avoid useless calls to `_prepare_data_table`

src/utils/connection_manager.py
```diff
@@ -0,0 +1,25 @@
+class Borg:
+    _shared_state = {}
+
+    def __init__(self):
+        self.__dict__ = self._shared_state
+
+
+class ConnectionManager(Borg):
+
```
chicco785, line 9: Basically this is a pool manager. Pools are not offered in the basic driver of crate and psql. It would be good to explore a proper db pool.

docker/docker-compose-dev.yml
```diff
@@ -48,6 +48,14 @@ services:
     volumes:
       - redisdata:/data
 
+  redis-commander:
+    image: rediscommander/redis-commander:latest
```
chicco785, line 52: so we can monitor what is cached

src/server/gconfig.py
```diff
@@ -66,7 +66,7 @@
 # Logging config section.
 #
 
-loglevel = 'debug'
+loglevel = 'error'
```
chicco785, line 69: debug is too verbose

.travis.yml
```diff
@@ -11,6 +11,7 @@ install:
 
 before_script:
   - pipenv install
+  - sudo service postgresql stop
```
chicco785, line 14: so we don't need weird ports

src/geocoding/location.py
```diff
@@ -24,7 +25,8 @@ def geometry_value(self) -> Optional[str]:
         return self._location.get(_VALUE_ATTR_NAME, None)
 
     def is_geojson(self):
-        return self.geometry_type() == _GEOJSON_TYPE
+        return self.geometry_type() == _GEOJSON_TYPE or \
+               self.geometry_type() == _GEOJSON_LD_TYPE
```
chicco785, line 29: to support proper management also of `GeoProperty` if NGSI-LD likey payload is sent

src/reporter/tests/test_notify.py
```diff
@@ -621,4 +621,112 @@ def test_no_value_with_type_for_attributes(service, notification):
     res_get = requests.get(url_new, headers=query_header(service))
     assert res_get.status_code == 200
     assert res_get.json()['values'][0] == None
+    delete_entity_type(service, notification['data'][0]['type'])
+
+
+@pytest.mark.parametrize("service", services)
+def test_issue_382(service, notification):
+    # entity with one Null value and no type
```
chicco785, line 629: can we store correctly 'relationship' type?

src/reporter/tests/test_notify.py
```diff
@@ -621,4 +621,112 @@ def test_no_value_with_type_for_attributes(service, notification):
...
+    # Give time for notification to be processed
+    time.sleep(SLEEP_TIME)
+    res_get = requests.get(url_new, headers=query_header(service))
+    assert res_get.status_code == 200
+    assert res_get.json()['values'][0] == 2
+    delete_entity_type(service, notification['data'][0]['type'])
+
+@pytest.mark.parametrize("service", services)
+def test_json_ld(service, notification):
+    # entity with one Null value and no type
```
chicco785, line 663: can we store correctly an NGSI-LD payload?

src/translators/crate.py
```diff
@@ -135,24 +144,45 @@ def _preprocess_values(self, e, table, col_names, fiware_servicepath):
...
-                        values.append([float(lon), float(lat)])
-                    else:
-                        values.append(e[cn]['value'])
+                    attr = original_attrs[cn][0]
+                    value = e[attr].get('value', None)
+                    if 'type' in e[attr] and e[attr]['type'] == 'geo:point':
+                        lat, lon = e[attr]['value'].split(',')
+                        value = [float(lon), float(lat)]
+                    elif 'type' in e[attr] and e[attr]['type'] == 'Property' \
+                         and 'value' in e[attr] \
```
chicco785, line 153: manage DateTime in NGSI-LD

src/translators/timescale.py
```diff
@@ -155,6 +185,16 @@ def _preprocess_values(self, e, table, col_names, fiware_servicepath):
                         mapped_value = str(ngsi_value)
                     elif mapped_type == PG_JSON_ARRAY:
                         mapped_value = pg8000.PGJsonb(ngsi_value)
+                    elif 'type' in e[attr] and e[attr]['type'] == 'Property' \
+                            and 'value' in e[attr] \
+                            and isinstance(e[attr]['value'], dict) \
+                            and '@type' in e[attr]['value'] \
+                            and e[attr]['value']['@type'] == 'DateTime':
+                        mapped_value = e[attr]['value']['@value']
```
chicco785, line 193: manage DateTime in NGSI-LD

src/translators/timescale.py
```diff
@@ -155,6 +185,16 @@ def _preprocess_values(self, e, table, col_names, fiware_servicepath):
...
+                    elif 'type' in e[attr] and e[attr][
+                            'type'] == 'Relationship':
+                        mapped_value = e[attr].get('value', None) or \
+                                       e[attr].get('object', None)
```
chicco785, line 197: manage Relationship in NGSI-LD

src/translators/crate.py
```diff
@@ -135,24 +144,45 @@ def _preprocess_values(self, e, table, col_names, fiware_servicepath):
...
+                         and isinstance(e[attr]['value'], dict) \
+                         and '@type' in e[attr]['value'] \
+                         and e[attr]['value']['@type'] == 'DateTime':
+                        value = e[attr]['value']['@value']
+                    elif 'type' in e[attr] and e[attr]['type'] == 'Relationship':
+                        value = e[attr].get('value', None) or \
+                                e[attr].get('object', None)
+
```
chicco785, line 161: manage Relationship in NGSI-LD

src/reporter/tests/test_notify.py
```diff
@@ -621,4 +621,112 @@ def test_no_value_with_type_for_attributes(service, notification):
...
+    notification['data'][0] = {
```
chicco785, line 664: @jason-fox @kzangeli is ngsi-ld payload representative enough? is there any tricky attribute type we should take into account? is DateTime the only date format that is specified via @value and @type ?

kzangeli: Attributes in NGSI-LD (at least right now) can have only three values in their type:
* Property
* Relationship
* GeoProperty

[ Two new attribute types are on their way: "LanguageProperty" and "VocabProperty", but those two are still under discussion in ETSI ISG CIM, so, never mind for now ]

A "Property" attribute has a "value" that can be of any JSON type.
A "Relationship" attribute has an "object" (instead of a "value") that is a string that is a valid URI
A "GeoProperty" has a "value" that is a GeoJSON. Eg: { "type": "Point", coordinates: [] }

There are a few special attributes, like observedAt, observationSpace etc. For that, see the [NGSI-LD spec](https://www.etsi.org/deliver/etsi_gs/CIM/001_099/009/01.03.01_60/gs_CIM009v010301p.pdf).

chicco785: thx @kzangeli

I think current pr would generally cover correctly
* Property
* Relationship
* GeoProperty

The issue is more for the case when the type of Property is not a native JSON Type. So far we implemented support for "DateTime" as by examples, but I could not find other similar not native types, to think of a generalisation. This is important especially if we want to provide backward and forward compatibility (as we are trying to do), where we meant the following:

* backward compatibility: data notified as NGSI-LD should be able to be retrieved as proper NGSI-v2
* forward compatibility: data notified as NGSI-v2 should be able to be retrieved as NGSI-LD

thx for any pointer you can provide to this tricky "custom" data type cases.

docs/manuals/admin/configuration.md
```diff
@@ -19,8 +19,11 @@ To configure QuantumLeap you can use the following environment variables:
 | `REDIS_HOST`       | Redis Host              |
 | `REDIS_PORT`       | Redis Port              |
 | `USE_GEOCODING`    | `True` or `False` enable or disable geocoding |
+| `DEFAULT_CACHE_TTL`| Time to live of metadata cache, default: 60 (seconds) |                              |
 | `QL_CONFIG`        | Pathname for tenant  configuration  |
-| `LOGLEVEL`         | define the log level for all services (`DEBUG`, `INFO`, `WARNING` , `ERROR`)      |
+| `QL_DEFAULT_DB`    | Default backend: `timescale` or `crate`  |
+| `USE_FLASK`        | `True` or `False` to use flask server or gunicorn. Default to `False`  |
```
chicco785, line 25: comment that is for development and testing, not for prod! (maybe rename)

docs/manuals/user/index.md
```diff
@@ -141,29 +141,46 @@ You need to make sure your NGSI entities are using the valid NGSI types for the
...
+|Text                | [text](https://www.postgresql.org/docs/current/datatype-character.html)                  | This is the default type if the provided NGSI Type is unsupported or wrong. |
+|StructuredValue     | [jsonb](https://www.postgresql.org/docs/current/datatype-json.html)                  |-|
+
+
+
 If the type of any of the received attributes is not present in the column
 *NGSI Type* of the previous table, the value of such attribute will be treated
 internally as a string. So, if you use `Float` for your attribute type (not
-valid), your attribute will be stored as a `string`.
+valid), your attribute will be stored as a `text`.
```
chicco785, line 181: if type not a NGSI type, the is inferred. with few exceptions to provide NGSI-LD forward compatibility

src/reporter/reporter.py
```diff
@@ -160,17 +152,23 @@ def notify():
...
+            # TODO in this way we return error for even if only one entity
+            #  is wrong
             return error, 400
         # Add TIME_INDEX attribute
         custom_index = request.headers.get(TIME_INDEX_HEADER_NAME, None)
         entity[TIME_INDEX_NAME] = \
             select_time_index_value_as_iso(custom_index, entity)
         # Add GEO-DATE if enabled
-        add_geodata(entity)
+        if not entity.get(LOCATION_ATTR_NAME, None):
```
chicco785, line 163: check if it's a geo:json or something

src/reporter/tests/test_notify.py
kzangeli: There's actually a fourth type of attribute/property - TemporalProperty, but that is kind of hidden.
Users aren't allowed to create a property of the type "TemporalProperty", but there is one "timestamp" that can be set by users.

NGSI-LD admits three different Temporal Properties:
* modifiedAt
* crateadAt
* observedAt

The two first are builtins, set by the broker and only observedAt is user-defined.
It's format is key-value basis:

"observedAt": "2020-11-15T10:00:00.000Z"

and, the broker will convert the timestamp into a floating point value.
This to be able to easily compare the timestamp with other timestamps (I imagine QL does the same).
Note though that the timestamps are converted back to ISO8601 form in queries and notifications.

About compound property values with "@type" and "@value", Orion-LD doesn't modify the value but keeps it as is.
This is actually something I've been thinking of taking up with ETSI ISG CIM but I haven't had any time for that still.

For example:
```
{
  "T1": {
    "type": "Property",
    "value": {
      "@type": "DateTime",
      "@value": "2020-..."
    }
}
```
T1 could be saved as a floating point for simpler and faster comparisons.

However, in the notifications that QL receives from the broker, T1 would always be converted back to its original form.
I believe this should be transparent to the subscriber (i.e. QL, Cygnus et al).

About compatibility questions, the easy answer is that:
* if NGSI-LD, simply compact entity types and attr names according to the current @context
* If NGSIv2, send the longnames without compaction.

The "not so easy" answer is that some features (datasetId) are simply not supported by NGSIv2 and there's probably no solution for this (well, if we really want to, there's always a way ...).

An attribute with datasetIds (or: multi-instances) are represented as an array in NGSI-LD, while that would provoke an error in NGSIv2:

"P1": [
  {
    "type": "Property",
    "value": 4,
    "datasetId": "urn:ngsi-ld:xxx:north"
  },
  {
    "type": "Property",
    "value": 9,
    "datasetId": "urn:ngsi-ld:xxx:south"
  }
]

My personal opinion is that this concept of notifying in NGSIv2 is a temporary thing that we may use with some constraints until we implement support for NGSI-LD.

BTW, the whole datasetId concept is quite complex and only partially implemented in Orion-LD as of today.

Right now I can't think of any other problems we may find.

chicco785: thx @kzangeli

c0c0n3 (approved):

c0c0n3:
@chicco785 good stuff! I think we can close #382 and #383 too, can't we?

PR #405, opened 2020-12-01
fix #404

### LINKED ISSUE #404: Can't use Timescale as default backend without setting QL_DEFAULT_DB
### Describe the bug
Configuring a default backend of `Timescale` in the QL config YAML file has no effect. When an HTTP request comes in with a tenant (`fiware-service` header) not in the config file, the selected backend always ends up being `Crate`. (In particular this is the case when the request has no `fiware-service` header or its value is the empty string.) The only way to make QL use `Timescale` as a default backend is to set the `QL_DEFAULT_DB` env var.

### To Reproduce
Steps to reproduce:

1. Create a YAML file with the content below:
```
tenants:
  t1:
    backend: Timescale
  t2:
    backend: Crate
  T3:
    backend: Timescale
default-backend: Timescale
```
2. Make the `QL_CONFIG` env var point to the file in (1).
3. Make sure the `QL_DEFAULT_DB` env var is **not** set.
4. Start both Crate and Timescale, configure QL with both backends, then start QL. 
5. Hit the `notify` endpoint with a new entity update.
6. Query Crate, you should be able to see the entity in (5).
7. Query Timescale, the entity in (5) shouldn't be there.

As a variant to the above, you can run Timescale only and you should be able to see an error when QL tries to process (5) since Crate isn't up---this is the scenario @DerLinne tested, see #391.

### Expected behavior
If `QL_DEFAULT_DB` isn't set and there's a config file with a `default-backend` stanza having a valid backend value `B`, `B` should be selected as a default backend for servicing incoming requests for unknown tenants.

### Additional context
It looks like this line is the culprit

- https://github.com/smartsdk/ngsi-timeseries-api/blob/master/src/translators/factory.py#L29

It sets the default backend to Crate if `QL_DEFAULT_DB` isn't set so then evaluating

    tenant_backend or env_backend or default_backend

becomes equivalent to evaluating

    tenant_backend or env_backend

chicco785:
## Proposed changes

see #404 

## Types of changes

## Checklist

      feature works
      downstream modules

## Further comments
separated default_backend computation logic for easy reusage

    "Contributing to QuantumLeap"

### DIFF
--- a/src/translators/factory.py
+++ b/src/translators/factory.py
@@ -24,11 +24,21 @@ def lookup_backend(fiware_service: str) -> MaybeString:
     config = cfg_reader.from_env_file(QL_CONFIG_ENV_VAR, defaults={})
     tenant_backend = maybe_string_match(config, 'tenants', fiware_service,
                                         'backend')
-    default_backend = maybe_string_match(config, 'default-backend')
 
-    env_backend = env_reader.read(StrVar(QL_DEFAULT_DB_ENV_VAR, CRATE_BACKEND))
+    return tenant_backend or default_backend()
 
-    return tenant_backend or env_backend or default_backend
+
+def default_backend() -> MaybeString:
+    cfg_reader = YamlReader(log=log().debug)
+    env_reader = EnvReader(log=log().debug)
+
+    config = cfg_reader.from_env_file(QL_CONFIG_ENV_VAR, defaults={})
+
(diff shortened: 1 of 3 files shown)


c0c0n3:
good stuff!!

c0c0n3 (dismissed):

c0c0n3 (approved):

Issue #445, opened 2021-02-04
split large data inserts


chicco785:
**Is your feature request related to a problem? Please describe.**

When a large set of data is received, if doing a single insert, crate bombs:

```
ERROR:app:Exception on /v2/notify [POST]
Traceback (most recent call last):
  File "/src/ngsi-timeseries-api/src/translators/sql_translator.py", line 295, in _insert_entity_rows
    self.cursor.executemany(stmt, rows)
  File "/usr/local/lib/python3.8/site-packages/crate/client/cursor.py", line 67, in executemany
    self.execute(sql, bulk_parameters=seq_of_parameters)
  File "/usr/local/lib/python3.8/site-packages/crate/client/cursor.py", line 53, in execute
    self._result = self.connection.client.sql(sql, parameters,
  File "/usr/local/lib/python3.8/site-packages/crate/client/http.py", line 331, in sql
    content = self._json_request('POST', self.path, data=data)
  File "/usr/local/lib/python3.8/site-packages/crate/client/http.py", line 458, in _json_request
    _raise_for_status(response)
  File "/usr/local/lib/python3.8/site-packages/crate/client/http.py", line 187, in _raise_for_status
    raise ProgrammingError(error.get('message', ''),
crate.client.exceptions.ProgrammingError: SQLActionException[CircuitBreakingException: [parent] Data too large, data for [<transport_request>] would be [1570329861/1.4gb], which is larger than the limit of [1530082099/1.4gb], usages [request=222986240/212.6mb, fielddata=0/0b, in_flight_requests=196118812/187mb, query=10660857/10.1mb, jobs_log=71309008/68mb, operations_log=67110264/64mb, accounting=10343710/9.8mb]]
```

**Describe the solution you'd like**

Split insert in smaller batches for multi notifications.

**Describe alternatives you've considered**

Fixing on the client side currently.

**Additional context**
N/A

c0c0n3:
This goes hand in hand with #193

chicco785:
@c0c0n3 fixing the inject to db with a sub batches should not be much pain, can you have a look? the issue occurred in prod on ekz

amotl:
Hi @c0c0n3 and @chicco785,

[...]

Issue #193, opened 2019-05-16
Batch processing of notifications


c0c0n3:
PR #191 added support to store multiple entities in notification messages---btw, kudos to @ohylli. But we still need to think about how to handle:

1. **Partial errors**. What to do if we encounter an error for some of the entities in the batch? e.g. what if 2 out of 20 entities fail validation? should we reject the whole batch (probably not a good idea!) or try saving the 18 that passed validation? What would the HTTP response look like in this case? How do we accumulate errors?
2. **Message size**. What to do with large batches---e.g. 100s of entities? Ideally we should process them in constant space to avoid eating up all available RAM. This applies both to the received HTTP message and the DB call we issue.

For now this is just a reminder that at some point in time we'll have to consider the points above. Also, we should probably split (1) and (2) in two separate PRs. 

Initial thoughts about a possible way to implement 

1. Leverage monoids/monads/monoidal functors as suggested [on stackoverflow](https://stackoverflow.com/questions/40427855/accumulating-errors-using-errort)
2. Use stream processing techniques to process a batch of batches of configurable size.

PR #450, opened 2021-02-10
Insert batches

c0c0n3:
## Proposed changes

This PR enables the splitting of the SQL rows to insert into batches. If the size of the list of rows the Translator has to insert is greater than a configured value `M`, the rows get split into smaller batches (lists), each having a size no greater than `M`, and each batch gets inserted separately, i.e. the Translator issues a separate SQL bulk insert for each batch. We do this since some backends (e.g. Crate) limit how much data you can shovel in a single SQL (bulk) insert statement---see #445 about it.
Splitting happens as explained in the notes below, using a cost function to compute how much data each row to insert holds in memory and a maximum batch size `M` (= cost in bytes) read from the env---see the notes below about configuration.

## Types of changes

## Checklist

- [x] Any dependent changes have been merged and published in downstream modules

### Splitting spec
We split a stream in batches so the cumulative cost of each batch is within a set cost goal. Given an input stream `s` and a cost function `c`, we want to produce a sequence of streams `b` such that joining the `b` streams yields `s` and, for each `b` stream of length `> 1`, mapping `c` to each element and summing the costs yields a value `≤ M`, where `M` is a configured cost goal. In symbols:

        (1)    s = b[0] + b[1] + ...
        (2)    b[k] = [x1, x2, ...] ⟹ M ≥ c(x1) + c(x2) + ...

Notice it can happen that to make batches satisfying (1) and (2), some `b[k]` contains just one element `x > M` since that doesn't violate (1) and (2).

### Implementation
We use Python streams to process data in constant space and linear time. Working with Python streams is anything but easy in my opinion so the implementation looks quite involved but the concept is fairly simple. In fact, for the mathematically inclined soul out there, the Python implementation is doing what this recursively defined function does (using Haskell-y syntax for lists), only in a more obscure way

    ϕ []          = []
    ϕ [x]         = [[x]]

    ϕ [x, y, ...] = [ x:t, u, ...]       if c(x) + Σ c(t[i]) ≤ M
    ϕ [x, y, ...] = [ [x], t, u, ...]    otherwise

       where  [t, u, ...] = ϕ [y, ...]

Notice this isn't a solution to #193 but is certainly one piece of the puzzle if we want to piece together a stream-based architecture. Why should we care? Well, even if we split the insert into batches, we still have two huge datasets in memory: the Python representation of the input NGSI JSON doc and its translation to tabular format. Ouch, not exactly a big-data friendly design. In an ideal world, the notify endpoint would work in constant space and linear time...

### Configuration
There's a new `INSERT_MAX_SIZE` env var to turn on the splitting into batches. If set, this variable limits how much data you can shovel in a single SQL bulk insert to a value `M`---see above for the details of how data gets split into batches of at most size `M`.  We read this variable in on each API call to the `notify` endpoint so it's sort of dynamic that way and will affect every later insert operation. Accepted values are sizes in bytes (B) or `2^10` multiples (KiB, MiB, GiB), e.g. `10 B`, `1.2 KiB`, `0.9 GiB`. (Technically, anything [bitmath][bitmath] can digest will do, e.g. `MB`, `kB`, and friends.) If the variable isn't set (or the set value isn't valid), the Translator processes SQL inserts normally without splitting data into batches.

[bitmath]: https://bitmath.readthedocs.io/

    "Contributing to QuantumLeap"


c0c0n3:
@chicco785 need your approval to merge :-)

Issue #444, opened 2021-02-04
support injection of data when an attribute is badly formatted


chicco785:
**Is your feature request related to a problem? Please describe.**

Currently, when an attribute is not correctly formatted, insertion fails totally.

**Describe the solution you'd like**

A more robust approach could be to check that attributes are of the correct type, if not removing them, and issuing a warning in the log.

**Describe alternatives you've considered**

N/A

**Additional context**
N/A

chicco785:
most probably we

PR #449, opened 2021-02-05
fix #446

chicco785:
## Proposed changes

- fix #446
- check health improvement using default backend for the check

## Types of changes

What types of changes does your code introduce to the project: _Put
an `x` in the boxes that apply_

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing
      functionality to not work as expected)

## Checklist

_Put an `x` in the boxes that apply. You can also fill these out after
creating the PR. If you're unsure about any of them, don't hesitate to
ask. We're here to help! This is simply a reminder of what we are going
to look for before merging your code._

- [x] I have read the [CONTRIBUTING][contrib] doc
- [x] I have signed the [CLA][cla]
- [ ] I have added tests that prove my fix is effective or that my
      feature works
- [ ] I have added necessary documentation (if appropriate)
- [ ] Any dependent changes have been merged and published in
      downstream modules

## Further comments

N/A

[cla]: https://fiware.github.io/contribution-requirements/individual-cla.pdf
    "FIWARE Individual Contributor License Agreement"
[contrib]: https://github.com/smartsdk/ngsi-timeseries-api/blob/master/CONTRIBUTING.md
    "Contributing to QuantumLeap"

PR #456, opened 2021-02-23
Fix 444

chicco785:
## Proposed changes

This PR aims at implementing #444 feature by introducing data casting functionalities:

* add data casting mechanism to increase resiliency due to mismatching data types
* uncastable values are dropped (None value is used as replacement), so that insert containing attributes that cannot be casted do not fail
* _preprocess_values is now harmonised between crate and timescale, only specific conversion are part of the specific translators
* as for crate, original ngsi is now stored as a string attribute, this to avoid issues with the fact that crate supports only a single column of type geo

The following table shows how the casting works through examples:

| Type          | Original value          | Stored value |
| ------------- |:-----------------------:| :------------|
| Number | 1.0 | 1.0 |
| Number | 1 | 1.0 |
| Number | True | None |
| Number | "1.0" | 1.0 |
| Number | "2017-06-19T11:46:45.00Z" | None |
| Integer | 1.0 | 1 |
| Integer | 1 | 1 |
| Integer | True | None |
| Integer | "1.0" | 1 |
| Integer | "2017-06-19T11:46:45.00Z" | None |
| DateTime | 1.0 | None |
| DateTime | 1 | None |
| DateTime | True | None |
| DateTime | "error" | None |
| DateTime | "2017-06-19T11:46:45.00Z" | "2017-06-19T11:46:45.000+00:00" |
| Text | 1.0 | "1.0" |
| Text | 1 | "1" |
| Text | True | "True" |
| Text | "1.0" | "1.0" |
| Text | "2017-06-19T11:46:45.00Z" | "2017-06-19T11:46:45.00Z" |
| Text | ["a", "b"] | "['a', 'b']" |
| Text | {"test": "test"} | "{'test': 'test'}" |
| StructuredValue | 1.0 | None |
| StructuredValue | 1 | None |
| StructuredValue | True | None |
| StructuredValue | "1.0" | None |
| StructuredValue | "2017-06-19T11:46:45.00Z" | None |
| StructuredValue | {"test": "test"} | {"test": "test"} |
| StructuredValue | ["a", "b"] | ["a", "b"] |
| Property | 1.0 | 1.0 |
| Property | 1 | 1 |
| Property | True | True |
| Property | "1.0" | "1.0" |
| Property | "2017-06-19T11:46:45.00Z" | "2017-06-19T11:46:45.000+00:00" |
| Property | {"test": "test"} | {"test": "test"} |
| Property | ["a", "b"] | ["a", "b"] |

In future more complex logics may be applied to casting, increasing number of values not stored as None.

## Types of changes

## Checklist

      feature works
      downstream modules

## Further comments

N/A

    "Contributing to QuantumLeap"


src/translators/sql_translator.py
```diff
@@ -455,8 +455,175 @@ def is_text(attr_type):
...
+        except ValueError:
+            logging.warning('{} cannot be cast to {} replaced with None'.format(attr['value'], attr['type']))
+            return None
+        else:
+            logging.warning('{} cannot be cast to {} replaced with None'.format(attr['value'], attr['type']))
+            return None
+
+    @staticmethod
+    def _ngsi_boolean_to_db(attr):
+        if isinstance(attr['value'], str) and attr['value'].lower() == 'true':
```
c0c0n3, line 577: should we strip whitespace? so e.g. `' true '`, `'true '`, etc. get converted to `True`? If so then we could just add a trailing call to `strip` as in `attr['value'].lower().strip()`.

src/translators/sql_translator.py
```diff
@@ -455,8 +455,175 @@ def is_text(attr_type):
...
+            return True
+        elif isinstance(attr['value'], str) \
```
c0c0n3, line 579: should we strip whitespace? so e.g. `' false '`, `'false '`, etc. get converted to `False`? If so then we could just add a trailing call to `strip` as in `attr['value'].lower().strip()`.

src/translators/sql_translator.py
```diff
@@ -455,8 +455,175 @@ def is_text(attr_type):
...
+        logging.warning('{} cannot be cast to {} replaced with None'.format(attr['value'], attr['type']))
+        return None
+
+    @staticmethod
+    def _ngsi_default_to_db(attr):
+        return attr.get('value', None)
+
+    @staticmethod
+    def _ngsi_ld_datetime_to_db(attr):
+        if SQLTranslator._is_iso_date(attr['value']['@value']):
```
c0c0n3, line 617: is this bomb proof? can we assume that this method only gets called if `attr is not None`, it is a `dict`, it has a `value` attribute which is also a `dict` (so not `None`) and this other dict contains a `@value` attribute. Gosh you gotta love Python :-)

src/translators/timescale.py
```diff
@@ -180,57 +179,30 @@ def _update_data_table(self, table_name, new_columns, fiware_service):
...
+    def _ngsi_slf_to_db(attr):
+        ast = SlfGeometry.build_from_ngsi_dict(attr)
+        return geocoding.slf.wktcodec.encode_as_wkt(ast, srid=4326)
+
+    @staticmethod
+    def _ngsi_structured_to_db(attr):
+        attr_v = attr.get('value', None)
+        if isinstance(attr_v, dict):
+            return pg8000.PGJsonb(attr_v)
+        logging.warning('{} cannot be cast to {} replaced with None'.format(attr['value'], attr['type']))
```
c0c0n3, line 196: Is this going to work in all cases? what if `attr` has no `value` attribute in it? ditto for `type`.

src/translators/timescale.py
```diff
@@ -180,57 +179,30 @@ def _update_data_table(self, table_name, new_columns, fiware_service):
...
+        return None
+
+    @staticmethod
+    def _ngsi_array_to_db(attr):
+        attr_v = attr.get('value', None)
+        if isinstance(attr_v, list):
+            return pg8000.PGJsonb(attr_v)
+        logging.warning('{} cannot be cast to {} replaced with None'.format(attr['value'], attr['type']))
```
c0c0n3, line 204: Is this going to work in all cases? what if attr has no value attribute in it? ditto for type.

src/translators/sql_translator.py
```diff
@@ -455,8 +455,175 @@ def is_text(attr_type):
...
+            return attr['value']['@value']
+        else:
+            logging.warning('{} cannot be cast to {} replaced with None'.format(attr['value']['@value'], attr['value']['@type']))
```
c0c0n3, line 620: Is this going to work in all cases? what if `attr` has no `@type` attribute in it?

src/translators/sql_translator.py
```diff
@@ -455,8 +455,175 @@ def is_text(attr_type):
...
+
+    @staticmethod
+    def _ngsi_array_to_db(attr):
+        raise NotImplementedError
+
+    @staticmethod
+    def _ngsi_text_to_db(attr):
+        if 'value' in attr and attr['value'] is not None:
+            return str(attr['value'])
+        logging.warning('{} cannot be cast to {} replaced with None'.format(attr['value'], attr['type']))
```
c0c0n3, line 608: Is this going to work in all cases? what if `attr` has no `type` attribute in it?

src/translators/sql_translator.py
```diff
@@ -455,8 +455,175 @@ def is_text(attr_type):
...
+                and attr['value'].lower() == 'false':
+            return False
+        elif isinstance(attr['value'], int) and attr['value'] == 1:
+            return True
+        elif isinstance(attr['value'], int) and attr['value'] == 0:
+            return False
+        elif isinstance(attr['value'], bool):
+            return attr['value']
+        else:
+            logging.warning('{} cannot be cast to {} replaced with None'.format(attr['value'], attr['type']))
```
c0c0n3, line 589: Is this going to work in all cases? what if `attr` has no `type` attribute in it?

src/translators/sql_translator.py
```diff
@@ -455,8 +455,175 @@ def is_text(attr_type):
...
+    def _ngsi_integer_to_db(attr):
+        try:
+            if isinstance(attr['value'], bool):
+                return None
+            elif isinstance(attr['value'], int):
+                return attr['value']
+            elif attr['value'] is not None:
+                return int(float(attr['value']))
+        except ValueError:
+            logging.warning('{} cannot be cast to {} replaced with None'.format(attr['value'], attr['type']))
```
c0c0n3, line 569: Is this going to work in all cases? what if `attr` has no `type` attribute in it?

src/translators/sql_translator.py
```diff
@@ -455,8 +455,175 @@ def is_text(attr_type):
...
+            return None
+        else:
+            logging.warning('{} cannot be cast to {} replaced with None'.format(attr['value'], attr['type']))
```
c0c0n3, line 572: Is this going to work in all cases? what if `attr` has no `type` attribute in it?

src/translators/sql_translator.py
```diff
@@ -455,8 +455,175 @@ def is_text(attr_type):
...
+            return None
+
+
+    @staticmethod
+    def _ngsi_datetime_to_db(attr):
+        if SQLTranslator._is_iso_date(attr['value']):
+            return attr['value']
+        else:
+            logging.warning('{} cannot be cast to {} replaced with None'.format(attr['value'], attr['type']))
```
c0c0n3, line 556: Is this going to work in all cases? what if `attr` has no `type` attribute in it?

src/translators/sql_translator.py
```diff
@@ -455,8 +455,175 @@ def is_text(attr_type):
...
+    def _ngsi_number_to_db(attr):
+        try:
+            if isinstance(attr['value'], bool):
+                return None
+            elif isinstance(attr['value'], float):
+                return attr['value']
+            elif attr['value'] is not None:
+                return float(attr['value'])
+        except ValueError:
+            logging.warning('{} cannot be cast to {} replaced with None'.format(attr['value'], attr['type']))
```
c0c0n3, line 544: Is this going to work in all cases? what if `attr` has no `type` attribute in it?

src/translators/sql_translator.py
```diff
@@ -455,8 +455,175 @@ def is_text(attr_type):
...
+            return None
+        else:
+            logging.warning('{} cannot be cast to {} replaced with None'.format(attr['value'], attr['type']))
```
c0c0n3, line 547: Is this going to work in all cases? what if `attr` has no `type` attribute in it?

src/translators/crate.py
```diff
@@ -122,41 +123,35 @@ def get_health(self):
...
+            return attr_v
+        logging.warning('{} cannot be cast to {} replaced with None'.format(attr['value'], attr['type']))
+        return None
+
+    @staticmethod
+    def _ngsi_array_to_db(attr):
+        attr_v = attr.get('value', None)
+        if isinstance(attr_v, list):
+            return attr_v
+        logging.warning('{} cannot be cast to {} replaced with None'.format(attr['value'], attr['type']))
```
c0c0n3, line 149: Is this going to work in all cases? what if `attr` has no `type` attribute in it?

src/translators/crate.py
```diff
@@ -122,41 +123,35 @@ def get_health(self):
...
+        if attr['type'] == 'geo:point':
+            lat, lon = attr['value'].split(',')
+            return [float(lon), float(lat)]
+
+    @staticmethod
+    def _ngsi_structured_to_db(attr):
+        attr_v = attr.get('value', None)
+        if isinstance(attr_v, dict):
+            return attr_v
+        logging.warning('{} cannot be cast to {} replaced with None'.format(attr['value'], attr['type']))
```
c0c0n3, line 141: Is this going to work in all cases? what if `attr` has no `type` attribute in it?

Pipfile
```diff
@@ -31,6 +31,7 @@ pytest-flask = "~=0.10"
 pytest = "~=3.0"
 pytest-cov = "~=2.7.1"
 coveralls = "~=2.0"
+lovely-pytest-docker = "*"
```
c0c0n3, line 34: I think we need to specify a version range here to avoid nasty surprises?

c0c0n3 (commented):
Great job! There's just some minor issues I picked up, but double check my comments perhaps they're not actual issues, since the code flow could be such that at the point in time when we call some of the methods those error conditions I'm talking about can't actually happen :-)

src/translators/sql_translator.py
chicco785: the method per se it is not, but it is in the chain with testing that data is supposed to be ngsi-ld time

Pipfile
chicco785: i think this is linked to the pr on top of which this is built, in case better fix there

src/translators/sql_translator.py
```diff
@@ -455,8 +455,175 @@ def is_text(attr_type):
...
+    @staticmethod
+    def _ngsi_geojson_to_db(attr):
         raise NotImplementedError
 
+    @staticmethod
+    def _ngsi_number_to_db(attr):
+        try:
+            if isinstance(attr['value'], bool):
+                return None
+            elif isinstance(attr['value'], float):
```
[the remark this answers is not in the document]
chicco785: have to check better, but I think I had to introduce because I noticed "value" manipulations

src/translators/sql_translator.py
```diff
@@ -455,8 +455,175 @@ def is_text(attr_type):
...
+        else:
+            logging.warning('{} cannot be cast to {} replaced with None'.format(attr['value'], attr['type']))
+            return None
+
+    @staticmethod
+    def _ngsi_integer_to_db(attr):
+        try:
+            if isinstance(attr['value'], bool):
+                return None
+            elif isinstance(attr['value'], int):
```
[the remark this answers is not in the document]
chicco785: see above for the manipulation (i will check with test if this actually occurs)

src/translators/sql_translator.py
chicco785: eheh no bombing in my tests, but you are right

src/translators/sql_translator.py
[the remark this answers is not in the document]
c0c0n3: then I'd leave it be, better to be safe than sorry :-)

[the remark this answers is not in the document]
c0c0n3: after all my suggestion would actually introduce an extra call to the `float` method which isn't actually needed if the value was a float to start with...

src/translators/sql_translator.py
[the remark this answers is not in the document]
c0c0n3: cool, then I think we it's best to leave it the way it is now? see my comments above too :-)

src/translators/sql_translator.py
c0c0n3: cool!

Pipfile
c0c0n3: yep, agreed!

src/cache/factory.py
```diff
@@ -47,7 +47,7 @@ def is_cache_available() -> bool:
         use geo-coding.
     """
     env = CacheEnvReader()
-    if env.redis_host():
+    if env.redis_host() and env.cache_queries():
```
c0c0n3, line 50: this change is already in #449, so perhaps we should merge #449 first and then double check what happens when rebasing, i.e. make sure nothing gets undone!

src/cache/factory.py
```diff
@@ -60,7 +60,7 @@ def get_cache() -> MaybeCache:
         object otherwise.
     """
     env = CacheEnvReader()
-    if is_cache_available() and env.cache_queries():
+    if is_cache_available():
```
c0c0n3, line 63: this change is already in #449, so perhaps we should merge #449 first and then double check what happens when rebasing, i.e. make sure nothing gets undone!

src/reporter/health.py
```diff
@@ -1,20 +1,25 @@
-import os
-
 from geocoding import geocoding
 from geocoding.factory import get_geo_cache, is_geo_coding_available
 from cache.factory import get_cache, is_cache_available
+from translators.factory import CRATE_BACKEND, TIMESCALE_BACKEND, \
```
c0c0n3, line 4: this change is already in #449, so perhaps we should merge #449 first and then double check what happens when rebasing, i.e. make sure nothing gets undone!

src/reporter/health.py
```diff
@@ -1,20 +1,25 @@
...
+    default_backend
 
 
-# TODO having now multiple backends, health check needs update
-def check_crate():
+def check_db(db=CRATE_BACKEND):
```
c0c0n3, line 8: this change is already in #449, so perhaps we should merge #449 first and then double check what happens when rebasing, i.e. make sure nothing gets undone!

src/reporter/health.py
```diff
@@ -1,20 +1,25 @@
...
     """
     crateDB is the default backend of QuantumLeap, so it is required by
     default.
     """
-    from translators.crate import CrateTranslatorInstance
-    with CrateTranslatorInstance() as trans:
-        crate_health = trans.get_health()
-        return crate_health
+    if db == CRATE_BACKEND:
```
c0c0n3, line 13: this change is already in #449, so perhaps we should merge #449 first and then double check what happens when rebasing, i.e. make sure nothing gets undone!

src/reporter/health.py
```diff
@@ -75,12 +80,15 @@ def get_health(with_geocoder=False):
         'status': 'pass'
     }
 
-    # Check crateDB (critical)
+    # Check defaultDB (critical)
+    db = default_backend().lower()
```
c0c0n3, line 84: this change is already in #449, so perhaps we should merge #449 first and then double check what happens when rebasing, i.e. make sure nothing gets undone!

src/reporter/health.py
```diff
@@ -75,12 +80,15 @@ def get_health(with_geocoder=False):
...
     try:
-        res = _check_critical(check_crate(), res, 'crateDB')
+        res = _check_critical(check_db(db), res, db)
```
c0c0n3, line 86: this change is already in #449, so perhaps we should merge #449 first and then double check what happens when rebasing, i.e. make sure nothing gets undone!

src/reporter/health.py
```diff
@@ -75,12 +80,15 @@ def get_health(with_geocoder=False):
...
     except Exception:
         res['status'] = 'fail'
-        res.setdefault('details', {})['crateDB'] = 'cannot reach crate'
+        res.setdefault('details', {})[db] = 'cannot reach ' + db
```
c0c0n3, line 89: this change is already in #449, so perhaps we should merge #449 first and then double check what happens when rebasing, i.e. make sure nothing gets undone!

src/reporter/tests/run_tests.sh
```diff
@@ -6,7 +6,6 @@ sleep 20
 cd ../../../
 pytest src/reporter/ \
        --cov-report= --cov-config=.coveragerc --cov-append --cov=src/ \
-       --ignore=src/reporter/tests/test_health.py \
```
c0c0n3, line 9: this change is already in #449, so perhaps we should merge #449 first and then double check what happens when rebasing, i.e. make sure nothing gets undone!

Pipfile
```diff
@@ -31,6 +31,7 @@ pytest-flask = "~=0.10"
 pytest = "~=3.0"
 pytest-cov = "~=2.7.1"
 coveralls = "~=2.0"
+lovely-pytest-docker = "~=0.2.0"
```
c0c0n3, line 34: this change is already in #449, so perhaps we should merge #449 first and then double check what happens when rebasing, i.e. make sure nothing gets undone! Also we should get rid of the `Pipfile.lock` in this PR and replace it with the one coming from #449 after rebasing, see my comments there.

c0c0n3 (commented):
cool, so it looks like it could be better to merge #449 before this PR, then rebase and double-check your change set? see my comments above....

chicco785:
@c0c0n3 rebase looks ok, and the build was ok as well

c0c0n3 (approved):
it looks like we have a clean merge now, thx @chicco785 !

PR #573, opened 2021-10-25
Fix for Issue#518

daminichopra:
## Proposed changes

Inserted a start check to verify that crate back off factor

## Types of changes

- [ x ] Bugfix (non-breaking change which fixes an issue)

## Checklist

- [ x ] I have read the [CONTRIBUTING][contrib] doc
- [ x ] I have signed the [CLA][cla]
      feature works
      downstream modules

## Further comments

Fix for [Issue](https://github.com/orchestracities/ngsi-timeseries-api/issues/518)

[cla]: https://raw.githubusercontent.com/orchestracities/ngsi-timeseries-api/master/individual-cla.pdf
    "Contributing to QuantumLeap"

### DIFF
--- a/src/translators/crate.py
+++ b/src/translators/crate.py
@@ -60,9 +60,6 @@ class CrateTranslator(sql_translator.SQLTranslator):
         # consecutive retries
         backoff_factor = EnvReader(log=logging.getLogger(__name__).debug) \
             .read(FloatRangeVar('CRATE_BACKOFF_FACTOR', 0.0))
-        import warnings
-        if backoff_factor > 120:
-            logging.warn("backoff_factor value should be less than 120")
         if self.connection is None:
             try:
                 self.connection = client.connect(


daminichopra:
Gentle Reminder!!

daminichopra:
Gentle Reminder!!

c0c0n3:
hi @daminichopra, thanks for this. I've just looked at the crate docs

- https://crate.io/docs/python/en/latest/connect.html#backoff-factor

where they explicitly mention the backoff factor should be between 0 and 120. So we need to output a config error not a warning. I suggest we change this PR as follows:

1. Implement a new [config reader](https://github.com/orchestracities/ngsi-timeseries-api/blob/master/src/utils/cfgreader.py) for float ranges, e.g. `class FloatRangeVar(Evar) ...`
    - `FloatRangeVar` gets instantiated with an interval `[a b]` where config values are expected to be, e.g. `[0, 120]` in your case.
    - on reading a config value `x`, `FloatRangeVar` raises a `ValueError` if `x` is not in `[a, b]`.
2. Simply use this new config reader in the translator, e.g. `backoff_factor = ... .read(FloatRangeVar('CRATE_BACKOFF_FACTOR', 0.0, 0, 120))`

This way we get to reuse the validation code in the future if we have similar env vars. What do you think? Does it make any sense?

src/utils/cfgreader.py
```diff
@@ -79,6 +79,19 @@ def _do_read(self, rep: str) -> float:
...
+    expected to be in range of 0 to 120 otherwise it will raise Value error.
+    """
+
+    def _do_read(self, rep: str) -> float:
+        backoff_factor = int(os.environ.get('CRATE_BACKOFF_FACTOR'))
+        if backoff_factor not in range(0,120):
+            raise ValueError('value out of range: {}'.format(backoff_factor))
+        return float(rep)
+
+
```
c0c0n3, line 94: Can we make this class more generic so it works for any range, not just the 0-120 range?
e.g.

```python
class FloatRangeVar(EVar):
   def __init__(var_name, default_value, mask_value, lo, hi):
    # super(var_name, default_value, mask_value)
    # ...check lo <= hi

   def _do_read(self, rep: str) -> float:
     value = float(rep)
     if value not in range(self.lo, self.hi):
       raise ValueError(...)
     return value
```

src/translators/crate.py
```diff
@@ -58,7 +59,7 @@ def setup(self):
         # Added backoff_factor for retry interval between attempt of
         # consecutive retries
         backoff_factor = EnvReader(log=logging.getLogger(__name__).debug) \
-            .read(FloatVar('CRATE_BACKOFF_FACTOR', 0.0))
+            .read(FloatRangeVar('CRATE_BACKOFF_FACTOR', 0.0))
```
c0c0n3, line 62: If we make `FloatRangeVar` generic (see comment below) then I think this line should be

```python
.read(FloatRangeVar(var_name='CRATE_BACKOFF_FACTOR', default_value=0.0, lo=0.0, hi=120.0)
```

c0c0n3 (commented):
@daminichopra thanks for your commits, if you could improve the code just slightly so we can reuse it for other range variables going forward---i.e. not just for the range `[0, 120]`. I sketched out in the review comments how we could do that. Also, we need to add unit tests for the new class `FloatRangeVar` similar to what we have already in `utils/tests/test_cfgreader`. Thank you soooo much!!

Issue #474, opened 2021-04-14
User/password support for CrateDB connection


IEMaestro:
**Is your feature request related to a problem? Please describe.**

Currently, QL version does not support user/password to connect to CrateDB. It uses the "user default" to connect and there is no ENV Variable to provide a password for this user. In case you want to enable the (external or not) access to CrateDB (e.g. to web-monitor the status of the historical DB) and you want to secure it (by using the CrateDB HBA option) this means that QL cannot get conected to the CrateBD. The only way to keep both (users authentication for SQL access/monitoring and QL access to Crate) requires to filter the IPs (internal) that can use the "crate (or defaut) superuser" with "trust" auth. mechanism. E.g. in a kubernetes env. where IPs may change and you cannot control all the assigned IPs, this means that all deployed pods within the cluster could potentially access CrateDB with superuser privileges.

**Describe the solution you'd like**

Support conection to CrateDB by specifiying a CrateDB user/password. If none are specified, use superuser "crate" (**which cannot be assigned a password, according Crate 4.1.x doc**) by default.  

**Describe alternatives you've considered**
I've check all the (official) doc corresponding to CrateDB 4.1.X. Only HBA is supported to enable authentication. CrateDB secure conf. (address parameter) only supports IP and IP masks (not hostnames) so I cannot restrict access to a given hostmane unless I know its (permanent) IP.

**Additional context**

Crate HBA doc:
https://crate.io/docs/crate/reference/en/4.5/admin/auth/

IEMaestro:
Thanks Andreas for your tracking and forwarding. But, correct me if I'm wrong: in case "crate" implements this "hostname" support, this would come with next Crate version (this is Crate 4.5.X) and, according to Quantum Leap installation guide, this only supports Crate 4.1.X, so, as far as I understand, this won't solve my issue, at least, not so far. Am I Right?
Regards!!

c0c0n3:
@IEMaestro thank you so much for bringing this up. Indeed backend authentication and authorisation are totally underdeveloped functionality areas in QuantumLeap. We'll definitely tackle this at some point in time and possibly support other authn/authz methods too. Even though we recognise the importance of having proper authn/authz in place, we can't push on with this issue at the moment since we've got another big chunk of work in the pipeline. But with luck, we might be able to work on this a couple of months from now. As always we'll gladly review PRs...mentioning it just in case anyone out there would like to contribute this feature.

amotl:
[...]

PR #574, opened 2021-10-26
Reporter tests timing issues

chicco785:
## Proposed changes

Fix tests failing due to cratedb data consolidation latency. While not being per se a bug of quantum leap,
with newly reduced resource on circle-ci, data consolidation takes a bit longer and this causes tests to fail.

## Types of changes

What types of changes does your code introduce to the project: _Put
an `x` in the boxes that apply_

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing
      functionality to not work as expected)

## Checklist

_Put an `x` in the boxes that apply. You can also fill these out after
creating the PR. If you're unsure about any of them, don't hesitate to
ask. We're here to help! This is simply a reminder of what we are going
to look for before merging your code._

- [x] I have read the [CONTRIBUTING][contrib] doc
- [x] I have signed the [CLA][cla]
- [x] I have added tests that prove my fix is effective or that my
      feature works
- [ ] I have added necessary documentation (if appropriate)
- [ ] Any dependent changes have been merged and published in
      downstream modules

## Further comments

N(/A

[cla]: https://raw.githubusercontent.com/orchestracities/ngsi-timeseries-api/master/individual-cla.pdf
    "Martel Open Source Software Individual Contributor License Agreement"
[contrib]: https://github.com/smartsdk/ngsi-timeseries-api/blob/master/CONTRIBUTING.md
    "Contributing to QuantumLeap"

PR #579, opened 2021-12-02
Support Crate DB Authentication

modularTaco:
## Proposed changes

Related to issue #474 

## Types of changes

## Checklist

      feature works
      downstream modules

## Further comments

Added a second crate docker container with authentication enabled and added this to all tests which are running on timescale an crate. This guarantees there are no missing permissions for the QL user in CrateDB.  
Note: Some tests failed randomly on my machine. Tried it on a clean Ubuntu VM and had the same problem. Running the tests again fixes it somestimes, but sometimes not. This was also the case before my changes. The error is always a `SchemaUnknownException` like `crate.client.exceptions.ProgrammingError: SchemaUnknownException[Schema 'mttest' unknown]` e.g. in test `test_cache_insert_with_tenant[crate]`. This happens with the `crate` as well as the `crate-auth` container.  
Hopefully the tests run fine in CircleCI.

[cla]: https://raw.githubusercontent.com/orchestracities/ngsi-timeseries-api/master/individual-cla.pdf
    "Contributing to QuantumLeap"

### DIFF
--- a/setup_dev_env.sh
+++ b/setup_dev_env.sh
@@ -28,6 +28,7 @@ export MONGO_HOST=$LH
 
 export QL_HOST=$LH
 export CRATE_HOST=$LH
+export CRATE_AUTH_HOST=$LH
 export POSTGRES_HOST=$LH
 
 export REDIS_HOST=$LH
--- a/setup_dev_env.sh
+++ b/setup_dev_env.sh
@@ -35,4 +35,4 @@ export REDIS_HOST=$LH
 
 echo "used ip: $LH"
 
-[[ "$SHELL" == "no" ]] || pipenv shell
+[[ "$PIPENV_SHELL" == "no" ]] || pipenv shell
--- a/setup_dev_env.sh
+++ b/setup_dev_env.sh
@@ -28,7 +28,6 @@ export MONGO_HOST=$LH
 
 export QL_HOST=$LH
 export CRATE_HOST=$LH
-export CRATE_AUTH_HOST=$LH
 export POSTGRES_HOST=$LH
 
 export REDIS_HOST=$LH
--- a/src/conftest.py
+++ b/src/conftest.py
@@ -13,9 +13,6 @@ QL_DEFAULT_DB = os.environ.get('QL_DEFAULT_DB', 'crate')
 
 CRATE_HOST = os.environ.get('CRATE_HOST', 'crate')
 CRATE_PORT = 4200
-
-CRATE_AUTH_HOST = os.environ.get('CRATE_AUTH_HOST', 'crate-auth')
-CRATE_AUTH_PORT = 4201
 CRATE_AUTH_USERNAME = "quantumleap"
 CRATE_AUTH_PASSWORD = "a_secret_password"
 
@@ -196,14 +193,14 @@ def crate_translator(clean_crate):
 @pytest.fixture()
 def clean_crate_auth():
     yield
-    do_clean_crate(crate_host=CRATE_AUTH_HOST,
-                   crate_port=CRATE_AUTH_PORT,
+    do_clean_crate(crate_host=CRATE_HOST,
+                   crate_port=CRATE_PORT,
                    crate_username=CRATE_AUTH_USERNAME,
                    crate_password=CRATE_AUTH_PASSWORD)
 
 @pytest.fixture()
 def crate_auth_translator(clean_crate_auth):
-    with crateTranslator(host=CRATE_AUTH_HOST, port=CRATE_AUTH_PORT, username=CRATE_AUTH_USERNAME, password=CRATE_AUTH_PASSWORD) as trans:
+    with crateTranslator(host=CRATE_HOST, port=CRATE_PORT, username=CRATE_AUTH_USERNAME, password=CRATE_AUTH_PASSWORD) as trans:
         yield trans
 
 
--- a/src/conftest.py
+++ b/src/conftest.py
@@ -14,6 +14,11 @@ QL_DEFAULT_DB = os.environ.get('QL_DEFAULT_DB', 'crate')
 CRATE_HOST = os.environ.get('CRATE_HOST', 'crate')
 CRATE_PORT = 4200
 
+CRATE_AUTH_HOST = os.environ.get('CRATE_AUTH_HOST', 'crate-auth')
+CRATE_AUTH_PORT = 4201
+CRATE_AUTH_USERNAME = "quantumleap"
+CRATE_AUTH_PASSWORD = "a_secret_password"
+
 POSTGRES_HOST = os.environ.get('POSTGRES_HOST', 'timescale')
 POSTGRES_PORT = 5432
 
@@ -27,6 +32,62 @@ ORION_HOST = os.environ.get('ORION_HOST', 'orion')
 ORION_PORT = os.environ.get('ORION_PORT', '1026')
 ORION_URL = "http://{}:{}/v2".format(ORION_HOST, ORION_PORT)
 
+
+from src.translators.crate import CrateTranslator
+
+class crateTranslator(CrateTranslator):
+
+    def insert(self, entities,
+                fiware_service=None, fiware_servicepath=None):
+        r = CrateTranslator.insert(self, entities,
+                                    fiware_service, fiware_servicepath)
+        self._refresh(set([e['type'] for e in entities]),
+                        fiware_service=fiware_service)
+        return r
+
+    def delete_entity(self, entity_id, entity_type=None,
+                        fiware_service=None, **kwargs):
+        r = CrateTranslator.delete_entity(self, entity_id, entity_type,
+                                            fiware_service=fiware_service,
+                                            **kwargs)
+        try:
+            self._refresh([entity_type], fiware_service=fiware_service)
+        except exceptions.ProgrammingError:
+            pass
+        return r
+
+    def delete_entities(self, entity_type=None, fiware_service=None,
+                        **kwargs):
+        r = CrateTranslator.delete_entities(self, entity_type,
+                                            fiware_service=fiware_service,
+                                            **kwargs)
+        try:
+            self._refresh([entity_type], fiware_service=fiware_service)
+        except exceptions.ProgrammingError:
+            pass
+        return r
+
+    def entity_types(self, fiware_service=None, **kwargs):
+        r = CrateTranslator.query_entity_types(
+            self, entity_type=None, fiware_service=fiware_service, **kwargs)
+        try:
+            self._refresh(r, fiware_service=fiware_service)
+        except exceptions.ProgrammingError:
+            pass
+        return r
+
+    def clean(self, fiware_service=None, **kwargs):
+        types = CrateTranslator.query_entity_types(
+            self, fiware_service=fiware_service, **kwargs)
+        if types:
+            for t in types:
+                CrateTranslator.drop_table(self, t,
+                                            fiware_service=fiware_service,
+                                            **kwargs)
+            try:
+                self._refresh(types, fiware_service=fiware_service)
+            except exceptions.ProgrammingError:
+                pass
 
 def do_clean_mongo():
     db_client = pm.MongoClient(MONGO_HOST, MONGO_PORT)
@@ -95,10 +156,10 @@ def orion_client():
     yield oc
 
 
-def do_clean_crate():
+def do_clean_crate(crate_host=CRATE_HOST, crate_port=CRATE_PORT, crate_username=None, crate_password=None):
     from crate import client
-    conn = client.connect(["{}:{}".format(CRATE_HOST, CRATE_PORT)],
-                          error_trace=True)
+    conn = client.connect(["{}:{}".format(crate_host, crate_port)],
+                          error_trace=True, username=crate_username, password=crate_password)
     cursor = conn.cursor()
 
     try:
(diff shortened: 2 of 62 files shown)


modularTaco:

chicco785:
@modularTaco i think it's better to rebase and merge this pr after @modularTaco  reworks of the test.
fyi the issue we are working on is essentially related to cratedb latency in consolidating data. depening on the hw capacities, test fails because we query the data before crate consolidates it (cf #574)

modularTaco:
> @modularTaco i think it's better to rebase and merge this pr after @modularTaco reworks of the test. fyi the issue we are working on is essentially related to cratedb latency in consolidating data. depening on the hw capacities, test fails because we query the data before crate consolidates it (cf #574)

Hey @chicco785 ,

thank you for the fast response!

You mean after @c0c0n3 reworks of the test?  
Makes totally sense to me. 

Do you have any more details on this issue (probably cratedb documention)? We experience this as well in the logs of our QL deployment from time to time.

chicco785:
> Hey @chicco785 ,
> 
> thank you for the fast response!
> 
> You mean after @c0c0n3 reworks of the test? Makes totally sense to me.

exactly

> 
> Do you have any more details on this issue (probably cratedb documentation)? We experience this as well in the logs of our QL deployment from time to time.

maybe @c0c0n3 can provide some references, but essentially this is related to the shared nothing and eventual consistent model adopted by crate.

chicco785:
@modularTaco can you sign the CLA?
I will now go through the code and give you my comments (if any)

setup_dev_env.sh
```diff
@@ -28,6 +28,7 @@ export MONGO_HOST=$LH
 
 export QL_HOST=$LH
 export CRATE_HOST=$LH
+export CRATE_AUTH_HOST=$LH
```
chicco785, line 31: i had a quick look at the whole changes:
i would not duplicate the configuration auth / no auth, this makes test longer for no reason.
so let's stick just add auth to the already deployed crate.

wdyt?

src/tests/docker-compose.yml
```diff
@@ -59,11 +60,24 @@ services:
       - PG_HOST=timescale
       - PG_PASS=*
 
+  crate-auth:
```
chicco785, line 63: se comment above

src/translators/crate.py
```diff
@@ -42,13 +42,15 @@
 class CrateTranslator(sql_translator.SQLTranslator):
     NGSI_TO_SQL = NGSI_TO_SQL
 
-    def __init__(self, host, port=4200, db_name="ngsi-tsdb"):
+    def __init__(self, host, port=4200, db_name="ngsi-tsdb", username=None, password=None):
```
chicco785, line 45: i don't recall where, but i experienced the case that i had to put 'crate' as username if using the driver with username to connect. if that the case, maybe better 'crate' as default value.

chicco785 (changes requested):
see my comments, and of course you need as well to rebase (should be straight forward)

modularTaco:
> @modularTaco can you sign the CLA? I will now go through the code and give you my comments (if any)

I think i have?

image

setup_dev_env.sh
modularTaco: I thought running the tests twice, with "normal" crate and with crate with authentication enabled, would be better to ensure the changes have no impact in environments with unauthenticated crates and to make it kind of easier to debug (if the test fails in crate and crate-auth, probably missing permissions is not the reason for this. if it works in crate but fails in crate-auth, it may be related to permissions). And additionally to make it easier to run the tests with older crate version (pre 4.5, where authentication was an enterprise feature), in that case the cratedb version env just needs to be set and the crate-auth tests deselected.

But i'm not sure if this is enough benefit to justify the longer test run

src/translators/crate.py
modularTaco: as far as i know (just looking in my memory, not in the docs) crate uses "crate" as username per default if the connection is done via the crate driver. if the connection uses the postgres driver "crate" needs to be used as username explicitly.

As the tests run fine with the (non-auth) crate container i think it works well.

i can set the default to crate instead of None and run the tests again, but i do not think that it'll make any difference

modularTaco:
> see my comments, and of course you need as well to rebase (should be straight forward)

Thank you for your quick response! Will rebase the branch tomorrow.

chicco785:
@modularTaco i forced the rebase to see if error get solved, unfortunately not.
again i don't think you need two different set of tests. you can simply add a specific test, in translator, by enabling and disabling authentication.

modularTaco:
> @modularTaco i forced the rebase to see if error get solved, unfortunately not. again i don't think you need two different set of tests. you can simply add a specific test, in translator, by enabling and disabling authentication.

Yes, unfortunately the error is still Relation unknown and happens in both targets, crate and crate-auth (so i guess it's still related to the tests and not to my changes).

What do you mean by "you can simply add a specific test, in translator, by enabling and disabling authentication."? To enable authentication i have to start crate with different params, so i need to shut crate down, change the startup params (or start another prepared container in the docker-compose file), wait until crate with auth has started, configure a dedicated user, password and permissions, run some kind of test (what do you suggest here? Just inserting some data?), shutdown crate with auth and start the normal crate again. If this is what you meant i can try to implement it later this week (and remove the duplicate tests for crate-auth)

chicco785:
> > @modularTaco i forced the rebase to see if error get solved, unfortunately not. again i don't think you need two different set of tests. you can simply add a specific test, in translator, by enabling and disabling authentication.
> 
> Yes, unfortunately the error is still Relation unknown and happens in both targets, crate and crate-auth (so i guess it's still related to the tests and not to my changes).
> 
> What do you mean by "you can simply add a specific test, in translator, by enabling and disabling authentication."? To enable authentication i have to start crate with different params, so i need to shut crate down, change the startup params (or start another prepared container in the docker-compose file), wait until crate with auth has started, configure a dedicated user, password and permissions, run some kind of test (what do you suggest here? Just inserting some data?), shutdown crate with auth and start the normal crate again. If this is what you meant i can try to implement it later this week (and remove the duplicate tests for crate-auth)

I think we can look into two options:
1. always use authenticated;
2. just test that auth works once in translator tests; in this case, hopefully, you are able to pass additional config via an env variable to the same docker service to enable auth

modularTaco:
hey @chicco785 ,

i removed all the duplicated tests and the second docker container.  
Instead i reconfigured the crate container to enable authentication and allow trust auth for everyone (so all tests can connect with user crate and no password) as follows:

```
    command: crate -Cauth.host_based.enabled=true
      -Cauth.host_based.config.0.user=crate
      -Cauth.host_based.config.0.address=_local_
      -Cauth.host_based.config.0.method=trust
      -Cauth.host_based.config.1.user=crate
      -Cauth.host_based.config.1.address=0.0.0.0/0
      -Cauth.host_based.config.1.method=trust
      -Cauth.host_based.config.99.method=password -Ccluster.name=democluster
      -Chttp.cors.enabled=true -Chttp.cors.allow-origin="*"
```

Afterwards i create a dedicated quantumleap user in `translators/tests/conftest.py` (as it is now, just in the `crate` instead of the `crate-auth` container).

This works fine, all translator tests pass.

In addition i've added a test to ensure the crate authentication works as expected:

```
@pytest.mark.parametrize("translator", [pytest.lazy_fixture('crate_auth_translator')], ids=["crate-auth"])
def test_auth_is_not_superuser(translator):
    stmt = "select current_user"
    translator.cursor.execute(stmt)
    res = translator.cursor.fetchall()
    current_user = res[0][0]
    assert current_user == 'quantumleap'
```

Running this with `pytest -k auth` (so this is the only test which will be run) works fine.

But if i run all tests, this is the only test which fails because of 

```
E       AssertionError: assert 'crate' == 'quantumleap'
E         - crate
E         + quantumleap
```

And at the moment i have no idea why it isn't using the correct pytest fixture in this case... Do you have any ideas here? 

I can push the changes to this branch if you wish, haven't done it yet because i didn't want to push commits with failing tests

chicco785:
> hey @chicco785 ,
> 
> i removed all the duplicated tests and the second docker container. Instead i reconfigured the crate container to enable authentication and allow trust auth for everyone (so all tests can connect with user crate and no password) as follows:
> 
> ```
>     command: crate -Cauth.host_based.enabled=true
>       -Cauth.host_based.config.0.user=crate
>       -Cauth.host_based.config.0.address=_local_
>       -Cauth.host_based.config.0.method=trust
>       -Cauth.host_based.config.1.user=crate
>       -Cauth.host_based.config.1.address=0.0.0.0/0
>       -Cauth.host_based.config.1.method=trust
>       -Cauth.host_based.config.99.method=password -Ccluster.name=democluster
>       -Chttp.cors.enabled=true -Chttp.cors.allow-origin="*"
> ```
> 
> Afterwards i create a dedicated quantumleap user in `translators/tests/conftest.py` (as it is now, just in the `crate` instead of the `crate-auth` container).
> 
> This works fine, all translator tests pass.
> 
> In addition i've added a test to ensure the crate authentication works as expected:
> 
> ```
> @pytest.mark.parametrize("translator", [pytest.lazy_fixture('crate_auth_translator')], ids=["crate-auth"])
> def test_auth_is_not_superuser(translator):
>     stmt = "select current_user"
>     translator.cursor.execute(stmt)
>     res = translator.cursor.fetchall()
>     current_user = res[0][0]
>     assert current_user == 'quantumleap'
> ```
> 
> Running this with `pytest -k auth` (so this is the only test which will be run) works fine.
> 
> But if i run all tests, this is the only test which fails because of
> 
> ```
> E       AssertionError: assert 'crate' == 'quantumleap'
> E         - crate
> E         + quantumleap
> ```
> 
> And at the moment i have no idea why it isn't using the correct pytest fixture in this case... Do you have any ideas here?
> 
> I can push the changes to this branch if you wish, haven't done it yet because i didn't want to push commits with failing tests

yes please update the branch, I will then review code :) if you add me to you project, I can also modify code as needed.

modularTaco:
> yes please update the branch, I will then review code :) if you add me to you project, I can also modify code as needed.

ok perfect, thank you

I've added you to the project and updated the branch

chicco785:
> But if i run all tests, this is the only test which fails because of
> 
> ```
> E       AssertionError: assert 'crate' == 'quantumleap'
> E         - crate
> E         + quantumleap
> ```

to speed up sql tasks, connection is cached (see https://github.com/modularTaco/ngsi-timeseries-api/blob/dev-crate-auth/src/translators/crate.py#L93), i solved by disposing the connection in the fixture (see https://github.com/modularTaco/ngsi-timeseries-api/blob/dev-crate-auth/src/conftest.py#L195). this may slow down tests, let's see, if that the case i will dispose the connection only in the auth tests.

now, i also refactored few things, and because of that integration tests brakes :) I will look into that tomorrow.

src/tests/docker-compose.yml
```diff
@@ -59,11 +59,17 @@ services:
       - PG_HOST=timescale
       - PG_PASS=*
 
-
   crate:
     image: crate:${CRATE_VERSION:-4.5.1}
-    command: crate -Cauth.host_based.enabled=false
-      -Ccluster.name=democluster -Chttp.cors.enabled=true -Chttp.cors.allow-origin="*"
+    command: crate -Cauth.host_based.enabled=true
```
chicco785, line 64: this is the reason why bc test are failing, ql 0.7.5 of course does not support auth

chicco785:
@modularTaco that's should be it :) time of test didn't increase so we can leave disposal where it is
have a look, if you don't have other things to change / add, we can merge it

chicco785 (approved):
LGTM

modularTaco:
Wow okay, a lot more than what i've expected

-
    > to speed up sql tasks, connection is cached 

    As i used different instances of the crate translator class, i thought the caching was related to each instance and therefore should not interfere. But i think i've missed the shared state of the connection manager here

- personally i'd prefer `CRATE_USERNAME` instead of `CRATE_DB_USERNAME`, as it refers to multiple databases/schemas in crate and not only to one, just like `CRATE_HOST` and `CRATE_PORT` do either. In case of timescale it looks like a parameter `POSTGRES_DB_NAME` is given as well, so the credentials will most likely be only valid for this specific database. Never used QL with timescale, so i'm not sure.

- just removed the TODO comment in the test, as you've already solved it


Thank you for the nice and extensive support! Looks very good to me now with your fixes

chicco785:
> Wow okay, a lot more than what i've expected
> 
> * > to speed up sql tasks, connection is cached
>   
>   
>   As i used different instances of the crate translator class, i thought the caching was related to each instance and therefore should not interfere. But i think i've missed the shared state of the connection manager here

given that at the time being we don't support different instances of crate or timescale, the connection manager is shared for db instance. this, of course, may change in future.

> * personally i'd prefer `CRATE_USERNAME` instead of `CRATE_DB_USERNAME`, as it refers to multiple databases/schemas in crate and not only to one, just like `CRATE_HOST` and `CRATE_PORT` do either. In case of timescale it looks like a parameter `POSTGRES_DB_NAME` is given as well, so the credentials will most likely be only valid for this specific database. Never used QL with timescale, so i'm not sure.

i think it works the same, the purpose of `DB` was to allow - if needed by some specific operations - to also have a superuser password.

> * just removed the TODO comment in the test, as you've already solved it
> 
> Thank you for the nice and extensive support! Looks very good to me now with your fixes

my pleasure, we welcome contributions, and do our best to guide contributors through them, of course depending on zillions of factors... but hey, that's life :)

src/conftest.py
```diff
@@ -128,23 +130,27 @@ def clean_crate():
...
+    crate_host = os.environ.get('CRATE_HOST', 'crate')
+    crate_port = 4200
+    crate_db_username = os.environ.get('CRATE_DB_USER', 'crate')
+    crate_db_password = os.environ.get('CRATE_DB_PASS', None)
 
     class Translator(CrateTranslator):
 
         def insert(self, entities,
-                   fiware_service=None, fiware_servicepath=None):
+                    fiware_service=None, fiware_servicepath=None):
```
Cerfoglg, line 142: Extra spaces here?

Cerfoglg (approved):
Looks good to me

src/conftest.py
chicco785: could be... next autopep8 will fix, unfortunately, on external pull requests we can't enforce autopep8 since we don't own the repo...

PR #593, opened 2021-12-14
Replace borg connection manager with thread safe db pool

daminichopra:
## Proposed changes

Does changes as per [references ](https://github.com/crate/crate-python/issues/393) 
Fix for [Issue ](https://github.com/orchestracities/ngsi-timeseries-api/issues/397)
## Types of changes

- [ x ] Bugfix (non-breaking change which fixes an issue)
- [ x ] New feature (non-breaking change which adds functionality)

## Checklist

- [ x ] I have read the [CONTRIBUTING][contrib] doc
- [ x ] I have signed the [CLA][cla]
      feature works
      downstream modules

## Further comments

- Replaced "borg" connection used in file connection_manager.py with  `CrateDB's SQLAlchemy dialect`

- Also, took references from [discussion](https://github.com/crate/crate-python/issues/393#issuecomment-738147528)

- used `create_engine` connection.

    "Contributing to QuantumLeap"

### DIFF
--- a/src/translators/base_translator.py
+++ b/src/translators/base_translator.py
@@ -17,7 +17,7 @@ class BaseTranslator(object):
         self.port = port
         self.db_name = db_name
 
-    def __enter__(self):
+    def __enter__(self, connection, query):
         self.setup()
         return self
 
--- a/src/translators/crate.py
+++ b/src/translators/crate.py
@@ -43,15 +43,15 @@ CRATE_TO_NGSI['string_array'] = 'Array'
 class CrateTranslator(sql_translator.SQLTranslator):
     NGSI_TO_SQL = NGSI_TO_SQL
 
-    def __init__(self, host, port=4200, db_name="ngsi-tsdb"):
-        super(CrateTranslator, self).__init__(host, port, db_name)
+    def __init__(self, connection, query , host, port=4200, db_name="ngsi-tsdb"):
+        super(CrateTranslator, self).__init__(host, connection, query , port, db_name)
         self.logger = logging.getLogger(__name__)
         self.dbCacheName = 'crate'
         self.ccm = None
         self.connection = None
         self.cursor = None
 
-    def setup(self):
+    def setup(self, connection, query):
         url = "{}:{}".format(self.host, self.port)
         self.ccm = ConnectionManager()
         self.connection = self.ccm.get_connection('crate')
--- a/src/translators/sql_translator.py
+++ b/src/translators/sql_translator.py
@@ -15,7 +15,8 @@ from uuid import uuid4
 
 from cache.factory import get_cache, is_cache_available
 from translators.insert_splitter import to_insert_batches
-from utils.connection_manager import Borg
+from utils.connection_manager import ConnectionManager
+
 # NGSI TYPES
 # Based on Orion output because official docs don't say much about these :(
 NGSI_DATETIME = 'DateTime'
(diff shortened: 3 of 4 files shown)


src/translators/base_translator.py
```diff
@@ -17,7 +17,7 @@ def __init__(self, host, port, db_name):
         self.port = port
         self.db_name = db_name
 
-    def __enter__(self):
```
chicco785, line 20: i don't think you need to change this method, the idea should be to replace the connection manager with sqlalchemy "create engine", but without changing interfaces

src/translators/crate.py
```diff
@@ -43,15 +43,15 @@
 class CrateTranslator(sql_translator.SQLTranslator):
     NGSI_TO_SQL = NGSI_TO_SQL
 
-    def __init__(self, host, port=4200, db_name="ngsi-tsdb"):
-        super(CrateTranslator, self).__init__(host, port, db_name)
+    def __init__(self, connection, query , host, port=4200, db_name="ngsi-tsdb"):
+        super(CrateTranslator, self).__init__(host, connection, query , port, db_name)
```
chicco785, line 47: see comment above

src/translators/crate.py
```diff
@@ -43,15 +43,15 @@
...
         self.logger = logging.getLogger(__name__)
         self.dbCacheName = 'crate'
         self.ccm = None
         self.connection = None
         self.cursor = None
 
-    def setup(self):
+    def setup(self, connection, query):
         url = "{}:{}".format(self.host, self.port)
```
chicco785, line 55: so, in here, you would use create_engine to get your connection using the proper parameters

src/translators/sql_translator.py
```diff
@@ -1781,11 +1782,11 @@ def _remove_from_cache(self, tenant_name, key):
                                     exc_info=True)
 
 
-class QueryCacheManager(Borg):
```
chicco785, line 1784: i think you can keep the borg here

src/translators/sql_translator.py
```diff
@@ -1781,11 +1782,11 @@ def _remove_from_cache(self, tenant_name, key):
...
+class QueryCacheManager(ConnectionManager):
     cache = None
 
-    def __init__(self):
-        super(QueryCacheManager, self).__init__()
+    def __init__(self, connection, query):
+        super(QueryCacheManager, self ).__init__( connection, query)
```
chicco785, line 1789: and leave this unchanged

src/utils/connection_manager.py
```diff
@@ -1,25 +1,67 @@
-class Borg:
-    _shared_state = {}
+from crate import client
```
chicco785, line 1: you should leave this class unchanged, because as said the idea is to replace directly in setup, secondly, you would need to proceed, so to say, in parallel and support both crate and timescale, easy if you keep this unchanged, if you touch this, you would need to adjust also timescale code, since it uses this code

src/translators/crate.py
```diff
@@ -43,15 +43,15 @@
...
         self.ccm = ConnectionManager()
```
chicco785, line 56: here could be an example (but don't take granted it works ;) ):

```
url = "crate://{}:{}".format(self.host, self.port)
        self.engine = sa.create_engine(url, connect_args={"pool_size": 10})
        self.connection = self.engine.connect()
```

of course it could be possible to create the two connection pools (the one for crate and the one for timescale), at the app start, and pass them, this would be probably the best option from a tech point, but you would need probably to change quite some code...

src/translators/sql_translator.py
daminichopra: updated as suggested

src/translators/crate.py
daminichopra: updated as suggested

daminichopra: @chicco785 , I have PR as suggested but some test are failing which are not shown up as in my local env.

PR #610, opened 2021-12-22
replace geocoder with geopy

chicco785:
the library geocoder is unmaintained since 4 years, so time to replace it. more, the new library supports async calls!

## Proposed changes

replace old library with new one

## Types of changes

## Checklist

      feature works
- [x] I have updated the [RELEASE_NOTES][RELEASE_NOTES.md]
      downstream modules

## Further comments

N/A

    "Contributing to QuantumLeap"

### DIFF
--- a/src/geocoding/geocoding.py
+++ b/src/geocoding/geocoding.py
@@ -122,7 +122,6 @@ def add_location(entity, raise_error=False, cache=None):
     try:
         info = asyncio.run(geocode(key))
 
-
     except Exception as e:
         logging.error(repr(e))
         if raise_error:
@@ -179,7 +178,11 @@ def _extract_most_accurate_osm_result(osm_response, geom_type):
     if len(results_of_specified_type) > 0:
         return results_of_specified_type[0].geojson
     elif geom_type == 'Point' and len(results) > 0:
-        return {'type': 'geo:point', 'value': "{}, {}".format(results[0].latitude, results[0].longitude)}
+        return {
+            'type': 'geo:point',
+            'value': "{}, {}".format(
+                results[0].latitude,
+                results[0].longitude)}
     return None
 
 
--- a/src/geocoding/geocoding.py
+++ b/src/geocoding/geocoding.py
@@ -299,8 +299,8 @@ def get_health():
     :return: dictionary with geocoder service health. ::see:: reporter.health.
     """
     try:
-        location = asyncio.run(geocode("New York city"))
-    except (requests.exceptions.RequestException, Exception) as e:
+        location = asyncio.run(geocode("New York City"))
+    except (Exception) as e:
         # geocoder docs say exception will be raised to the caller
         time = datetime.now().isoformat()
         output = "{}".format(e)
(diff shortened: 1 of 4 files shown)


src/geocoding/geocoding.py
```diff
@@ -290,40 +285,31 @@ def is_valid_address(street, number, locality, region, country):
...
+        return info
 
 
 def get_health():
     """
     :return: dictionary with geocoder service health. ::see:: reporter.health.
     """
     try:
-        g = geocoder.osm("New York city", maxRows=1)
+        location = asyncio.run(geocode("New York city"))
```
Cerfoglg, line 302: New York City?

Cerfoglg (approved):
Overall, LGTM

src/geocoding/geocoding.py
```diff
@@ -290,40 +285,31 @@ def is_valid_address(street, number, locality, region, country):
...
-    r = requests.get(url, params=params)
-    if r.ok:
-        return r.json().get('geojson', None)
+async def geocode(key: str):
+    async with Nominatim(
+            user_agent="quantumleap",
+            adapter_factory=AioHTTPAdapter,
+    ) as geolocator:
+        info = await geolocator.geocode(key, geometry='geojson', exactly_one=False, limit=5, timeout=30)
+        return info
```
c0c0n3, line 294: @chicco785 you sure you want to swap out the sync API for the async one? I'm not convinced in general we're going to get major performance benefits without changing the current QL architecture. Isn't this code just a case in point? i.e. we're using the async API but then in practice the await statement makes the call synchronous just like we had it before. Am I missing something here?

chicco785: @c0c0n3 i think it would be ideal to go async with all external calls, this would be just a first step

c0c0n3: Yeah, I like the idea, but I'm a bit scared about the effort we'll have to put in. Like I said, I don't think we're gonna be able to improve performance significantly without changing the architecture. You can read about why I think that [over here](https://github.com/orchestracities/ngsi-timeseries-api/wiki/No-async-free-Lunch) and also change this code accordingly if you have time, otherwise we can just leave it be, we can refactor later.

chicco785: @c0c0n3 given that the introduced overhead maybe negligible (in your measurements is 0.2%), i think a gradual approach makes sense. also in this specific case, considering your comments about avoiding duplicated calls, we already use caching, i.e. we call the geocode only if the values are not in the cache.

in general, I believe it would be good to think if there are things that can be run in parallel, but that only as a second step. for example, instead of adding the geo-resolution of data immediately when injecting data, we could do that in a second step. in general this could apply to any secondary data manipulation, and could be also the way to parallelize inserts of batches.

c0c0n3: sure, if there's any easy wins we should go for it! re: the overhead, there's no overhead if we go async---sorry if that wasn't clear in the wiki page I linked, I wrote that too quickly I guess. The point is the code we have here should change if you want to the `geocode` calls to run concurrently at the moment they get serialised---e.g. by using `gather` as explained in the wiki page.

c0c0n3 (approved):

Issue #624, opened 2022-02-11
DockerHub image proliferation


c0c0n3:
### Describe the bug
There's one helluva lot QL images on DockerHub at the moment, which could be confusing for our users. Was this intentional? Also unlike previous versions there's no `0.8.3` tag, but `release-0.8.3` instead. That intentional too?

### Expected behavior
Double check what's going on w/ CircleCI and if we're supposed to publish an image on every PR merge.

### Additional context
Browse to: https://hub.docker.com/r/orchestracities/quantumleap/tags

chicco785:
for docker images, we are using github actions, no circle ci:
https://github.com/orchestracities/ngsi-timeseries-api/blob/master/.github/workflows/docker.yml

chicco785:
i think the issue could be that for whatever reason the script takes the image tag from the github release name and not from the github tag...

c0c0n3:
oh forgot to mention. `latest` is 7 months old

- https://hub.docker.com/r/orchestracities/quantumleap/tags?page=1&name=latest

so if someone pulls QL w/o specifying a tag, they'll get an old image.

c0c0n3:
yeh, it could be. not sure how it actually works...

chicco785:
https://github.com/orchestracities/ngsi-timeseries-api/runs/4546570621?check_suite_focus=true apparently the problem was that the build failed

PR #625, opened 2022-02-11
fix issues with ci worfklow for docker

chicco785:
## Proposed changes

The github action is not correctly generating tags for the docker images. e.g. instead of generating 0.8.3 tag, generated release-0.8.3, and didn't tag it also as "latest".

## Types of changes

## Checklist

- [ ] Any dependent changes have been merged and published in downstream modules

## Further comments

N/A

    "Contributing to QuantumLeap"
    "QuantumLeap Release Notes"

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -32,7 +32,6 @@ jobs:
             type=semver,pattern={{version}}
             type=semver,pattern={{major}}.{{minor}}
             type=semver,pattern={{major}}
-            type=ref,event=branch
       -
         name: Login to DockerHub
         if: github.event_name != 'pull_request'
@@ -68,7 +67,6 @@ jobs:
             type=semver,pattern={{version}}
             type=semver,pattern={{major}}.{{minor}}
             type=semver,pattern={{major}}
-            type=ref,event=branch
       -
         name: Login to DockerHub
         if: github.event_name != 'pull_request'
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -34,7 +34,7 @@ jobs:
             type=semver,pattern={{major}}
       -
         name: Login to DockerHub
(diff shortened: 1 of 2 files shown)


chicco785:
@c0c0n3 checking at the actions outcomes, it seems that actually the issue was that the build failed.
anyhow, i have removed the creation of images linked to branches.

fix #624

chicco785:
> great job! just one question. shouldn't this yaml at the top of the file
> 
> ```yaml
> on:
>   push:
>     branches:
>       - '**'
>     tags:
>       - '*.*.*'
>       - '*.*'
>   pull_request:
>     branches:
>       - 'master'
> ```
> 
> rather be something like
> 
> ```yaml
> on: [release]
> ```
> 
> ?
> 

The workflow should run all the times to verify that a docker build actually works. That's why is not only on release.

> The GitHub docs are quite terse (https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release) but if I think that should trigger the publishing of the Docker image only when we cut a release, i.e. when we publish a release on GitHub:
> 
> * https://github.com/orchestracities/ngsi-timeseries-api/releases
> 
> But if we do this then we don't have the convenience of publishing images on a merge into master without having to cut a release. What do you think is best
> 
> 1. Only push a docker image to DockerHub when we cut a release---possibly less images for our users to browse.
> 2. Also push when merging into master---convenient and we used it in the past.
> 
> I'm inclined to think (2) would be best, since users can easily filter out images on DockerHub that don't follow the semver pattern. If you agree, then I guess we can just merge this PR?

I think (and we are going to see now...) a new 'edge' tagged release is going to be created every time we merge something into master.

chicco785:
> I think (and we are going to see now...) a new 'edge' tagged release is going to be created every time we merge something into master.

it turned out to be a wrong assumption... so I will have to fix

PR #646, opened 2022-05-06
Update the connection data from env on translator creation

wistefan:
## Proposed changes

The change reads out the enviornment variables for connection data on translator connection. This is required, because else the health-check fails until the first real call is made to the api. This leads to a deadlock in k8s environments, since no call is possible before its healthy.

## Types of changes

- [ ] Maintenance (update of libraries or other dependences)

## Checklist

- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] Any dependent changes have been merged and published in downstream modules

## Further comments

    "Contributing to QuantumLeap"
    "QuantumLeap Release Notes"

### DIFF
--- a/src/translators/timescale.py
+++ b/src/translators/timescale.py
@@ -82,7 +82,6 @@ class PostgresTranslator(sql_translator.SQLTranslator):
     def __init__(self, conn_data=PostgresConnectionData()):
         super(PostgresTranslator, self).__init__(
             conn_data.host, conn_data.port, conn_data.db_name)
-        conn_data.read_env()
         self.logger = logging.getLogger(__name__)
         self.db_user = conn_data.db_user
         self.db_pass = conn_data.db_pass
--- a/src/translators/timescale.py
+++ b/src/translators/timescale.py
@@ -82,6 +82,7 @@ class PostgresTranslator(sql_translator.SQLTranslator):
     def __init__(self, conn_data=PostgresConnectionData()):
         super(PostgresTranslator, self).__init__(
             conn_data.host, conn_data.port, conn_data.db_name)
+        conn_data.read_env()
         self.logger = logging.getLogger(__name__)
         self.db_user = conn_data.db_user
         self.db_pass = conn_data.db_pass
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -13,6 +13,7 @@
   will be replaced with /. This affects only users that manually injected
   data, since Orion, assume / when no servicePath is passed.
 - Added more test cases for Aggregation (#499)
+- fix translator initialization
 
 ### Bug fixes
 
--- a/docs/manuals/admin/configuration.md
+++ b/docs/manuals/admin/configuration.md
@@ -208,13 +208,13 @@ configuration:
 ```yaml
 tenants:
     t1:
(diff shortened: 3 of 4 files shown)


wistefan:

src/translators/timescale.py
```diff
@@ -82,6 +82,7 @@ class PostgresTranslator(sql_translator.SQLTranslator):
     def __init__(self, conn_data=PostgresConnectionData()):
         super(PostgresTranslator, self).__init__(
             conn_data.host, conn_data.port, conn_data.db_name)
+        conn_data.read_env()
```
c0c0n3, line 85: Is this actually going to work? The connection data gets read after class init! Did you test this? :-) Not a Python expert myself, but maybe we can move this statement before the call to `super`?

On the other hand, if Python moans about super not being the first call in the ctor, then you can zap this statement (i.e. `conn_data.read_env()`) and fix the health endpoint instead. In fact, the PostgressTranslator ctor wasn't meant to be called directly but rather acquired through a [context manager](https://github.com/wistefan/ngsi-timeseries-api/blob/6e2e11d09405ac9488e3dd676ac8a38f49830c8d/src/translators/timescale.py#L314). So the code in the [health endpoint](https://github.com/wistefan/ngsi-timeseries-api/blob/master/src/reporter/health.py#L18-L22) should've rather been

```python
    if db == TIMESCALE_BACKEND:
        from translators.timescale import postgres_translator_instance
        with postgres_translator_instance() as trans:
            health = trans.get_health()
            return health
```

As you've probably noticed our code is quite tangled and is in need of refactoring :-)

c0c0n3:
hi @wistefan and welcome on board! Thanks so much for your contrib! I've done a first review, see my comments there. Basically if you've tested the fix we could keep it---I'm just not sure about the read_env call having the effect you're hoping for, so if you could please double check.

Thanks a million!!

src/translators/timescale.py
chicco785: @wistefan  i tend to think this change is not needed, and would like to know why you want to inject it, i.e. what's is the issue you are facing.

Context Manager is used to in the factory, so in the normal processing of incoming requests, configuration is correctly loaded:
https://github.com/orchestracities/ngsi-timeseries-api/blob/master/src/translators/factory.py#L53

chicco785: ok, i got the issue from andrea's comment, so the correct fix is what he proposed. @wistefan can you apply it?

thx

chicco785: with the fix, you won't need `conn_data.read_env()` anymore, can you remove?
many thx

wistefan: Good morning, thanks for the quick reaction. I tested it and it worked. But I prefer your solution, that looks much cleaner. Applied in c39f8225d2876bf68b874eaf867b8f65f477f1ef

chicco785: @wistefan i saw that you applied the solution, but the code in this pr still include `conn_data.read_env()`, can you remove it?

wistefan: Your right, sorry forgot to push that part. Fixed in 7bcc862656b949888d8f120271caab5b850b1670

c0c0n3 (approved):

Issue #697, opened 2022-10-19
Improve error handling


pooja1pathak:
Resolve TODO at: 
https://github.com/orchestracities/ngsi-timeseries-api/blob/6ecedfe5e81e346fc02e3cbd0bc86e6c20fab25a/src/translators/sql_translator.py#L371

PR #703, opened 2022-11-21
resolved error handling in sql_translator #697

NEC-Vishal:
## Proposed changes
#697 improve error handling in sql_translator.py

## Types of changes

- [x] Maintenance (update of libraries or other dependences)

## Checklist

- [x] Any dependent changes have been merged and published in downstream modules

## Further comments

If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you did and what alternatives you considered, etc...

    "Contributing to QuantumLeap"
    "QuantumLeap Release Notes"

### DIFF
--- a/src/translators/sql_translator.py
+++ b/src/translators/sql_translator.py
@@ -362,15 +362,13 @@ class SQLTranslator(base_translator.BaseTranslator):
         stmt = f"insert into {table_name} ({col_list}) values ({placeholders})"
         try:
             start_time = datetime.now()
+            res_list = []
 
             for batch in to_insert_batches(rows):
                 res = self.cursor.executemany(stmt, batch)
-                # new version of crate does not bomb anymore when
-                # something goes wrong in multi entries
-                # simply it returns -2 for each row that have an issue
-                # TODO: improve error handling.
-                # using batches, we don't need to fail the whole set
-                # but only failing batches.
+                res_list.append(res)
+        
+            for res in res_list:
                 if isinstance(res, list):
                     for i in range(len(res)):
                         if res[i]['rowcount'] < 0:


src/translators/sql_translator.py
```diff
@@ -362,15 +362,12 @@ def _insert_entity_rows(self, table_name: str, col_names: List[str],
...
             for batch in to_insert_batches(rows):
                 res = self.cursor.executemany(stmt, batch)
-                # new version of crate does not bomb anymore when
-                # something goes wrong in multi entries
-                # simply it returns -2 for each row that have an issue
-                # TODO: improve error handling.
-                # using batches, we don't need to fail the whole set
-                # but only failing batches.
+                res_list.append(res)
+            for res in res_list:
```
c0c0n3, line 370: This approach will only work for Crate, not Timescale which I think throws an exception instead. In general, what to do here will depend on the database at hand, so we should actually push this code down to the translator subclasses.

c0c0n3 (commented):
I initially thought your approach was great, but then I realised that unfortunately b/c of the spaghetti code we've got, we'll need a more involved approach.

c0c0n3:
@NEC-Vishal this TODO is harder to get right than I initially thought.

There are two problems with the code as it stands in the master branch:

1. If any batch fails, all the following ones won't be inserted. This is what the TODO mentioned.
2. The error handling block below the TODO is Crate-specific.

I propose we do the following. (Please feel free to propose a different approach if you don't agree.)

First off,  we need an abstract method in the `SQLTranslator` class to insert each and every batch regardless of whether any previous inserts failed.

```python
class SQLTranslator:

    def _insert_row_batches(stmt: str, batches: [[Tuple]]):
        raise NotImplementedError

    def _insert_entity_rows(...
        ...
        try:
            batches = to_insert_batches(rows)
            self._insert_row_batches(stmt, batches)
        except Exception as e:
            ...
```

In the case of one or multiple batches failing, `_insert_row_batches` will throw an exception. If there are multiple failures, they all get wrapped up in a single exception. This is obviously far from optimal, but at least the existing error handling code in `_insert_entity_rows` will keep on working as expected, in particular the last-ditch attempt to save entity data in the original format.

Then this should be the Crate-specific implementation.

```python
class CrateTranslator:

    def _insert_row_batches(stmt: str, batches: [[Tuple]]):
        has_error = False
        for batch in batches:
            res = self.cursor.executemany(stmt, batch)
            # recent Crate versions don't bomb out anymore when
            # something goes wrong with a multi-row insert, instead
            # the driver returns -2 for each row that has an issue
            if isinstance(res, list):
                for i in range(len(res)):
                    if res[i]['rowcount'] < 0:
                        has_error = True
        if has_error:
            raise Exception('An insert failed')
```

Whereas for Timescale, we could have something like this

```python
class PostgresTranslator:

    def _insert_row_batches(stmt: str, batches: [[Tuple]]):
        error = None
        for batch in batches:
            try:
                self.cursor.executemany(stmt, batch)
            except Exception as e:
                error = e
        if error:
            raise error
```

c0c0n3:
@NEC-Vishal please keep in mind I haven't tested any of this and the pseudo-code I sketched out earlier may need some tweaking. But the biggest chunk of work in this PR will be the testing. This is a critical section of the code base, so you'll need to implement test cases to cover all possible scenarios---no failure, partial failure, total failure. Both for Crate and Timescale...

